PR #33, opened 2022-10-19
Okhttp multi interceptor issue

LikeTheSalad:
Okhttp [interceptors](https://square.github.io/okhttp/features/interceptors/) can be many, they can change the request and they are chained. In order to change a request, the interceptor must first [clone the existing one in the form of a Request builder](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-request/new-builder/) that can be modified, so a new instance of Request is created and passed down to the next elements in the chain. 

Interceptors are always called in the order they were set, so only the first one is guaranteed to have access to the "original" request object, which is the one mapped to the OTel context. Therefore is is important that our Interceptor is the first one in the chain, so that it can find the parent OTel context, and propagate it. The code changes in this PR make sure that our interceptor is always the first one in the chain.


android-instrumentation/src/main/java/co/elastic/apm/android/instrumentation/okhttp/client/OkHttpClientAdvice.java
```diff
@@ -34,6 +34,6 @@ public static void enter(@Advice.Argument(0) OkHttpClient.Builder builder) {
         OkHttpContextStore contextStore = new OkHttpContextStore();
         OtelOkHttpEventListener.Factory otelFactory = new OtelOkHttpEventListener.Factory(contextStore);
         builder.eventListenerFactory(new CompositeEventListenerFactory(otelFactory, builder.getEventListenerFactory$okhttp()));
-        builder.addInterceptor(new OtelOkHttpInterceptor(contextStore));
+        builder.interceptors().add(0, new OtelOkHttpInterceptor(contextStore));
```
LikeTheSalad, line 37: This is the only production code change, the rest is some testing setup needed to verify this change.

felixbarny:
Have you considered attaching the context via a [`okhttp3.Request.Builder#tag(java.lang.Class<? super T>, T)`](https://square.github.io/okhttp/3.x/okhttp/okhttp3/Request.Builder.html#tag-java.lang.Class-T-)? If that works, we don't depend on the order of the interceptors which feels less brittle.

This PR contains a bunch of changes that are unrelated to the interceptor issue. I find it easier to review changes in isolation. Not a big issue though, just something I noticed.

android-instrumentation/src/main/java/co/elastic/apm/android/instrumentation/okhttp/client/OkHttpClientAdvice.java
felixbarny: I missed this when I wrote my comment. Now I understand why this PR contains seemingly unrelated changes.

LikeTheSalad:
That's very interesting, I didn't know that the Request object could store tags tbh. It could be very handy, however currently we're mapping the current context to a request [here](https://github.com/elastic/apm-agent-android/blob/main/android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/http/impl/okhttp/OtelOkHttpEventListener.java#L67) - This is needed because the `callStart` method is called before OkHttp launches its secondary threads, so we can still get the current context there. But the EventListener is not part of the interceptors chain, so it cannot modify the Request and pass it along to the interceptors, so even if we created a new Request in `callStart` and set the tag, interceptors will still get the original one. So if we wanted to use tags, we'd still need to make sure that we have some interceptor early in the chain that adds the tag so that further interceptors can get it. But interceptors are called in some secondary thread created by OkHttp where the OTel context is no longer available 😅

android-instrumentation/src/main/java/co/elastic/apm/android/instrumentation/okhttp/client/OkHttpClientAdvice.java
LikeTheSalad: No worries, hopefully this setup will be needed only once, further tests will just use it so this kind of big changes should only happen once ;)

Issue #20, opened 2022-09-08
[META] Improve onboarding experience for Android devs


LikeTheSalad:
- [x] Provide utilities for good onboarding experience (i.e. configuration and setup of the agent on a "happy path")
- [x] Write getting started docs - https://github.com/elastic/apm-agent-android/issues/25

https://www.elastic.co/guide/en/apm/agent/java/current/configuration.html

Issue #16, opened 2022-08-29
Make setting Elastic cloud's auth token more secure


LikeTheSalad:
Currently in order to connect the Android agent to an Elastic cloud instance, users need to pass their instance's endpoint + its auth token. Currently their auth token is passed in plain text, we need to figure ways to make it more secure.

PR #34, opened 2022-10-19
Agent configuration 

LikeTheSalad:
Related to #20 and #16 

Added gradle configuration that looks like the following:

```groovy
elasticApm {
  serverUrl = "http://..." // mandatory
  serverToken = "someToken" // optional, could be set using environment variables, defaults to null.
  serviceName = "My App" // optional, defaults to the Android package name.
  serviceVersion = "1.0.0" // optional, defaults to "android.defaultConfig.versionName".
}
```

Thanks to the addition of the above config, now it's optional to pass the `Connectivity` config to the Elastic agent initialization. So the simplest use case is now:

`ElasticApmAgent.initialize(AndroidContext)` (Used to be ElasticApmAgent.initialize(AndroidContext, Connectivity))


trentm:
@LikeTheSalad Possibly ignorant question: Shouldn't `serverToken` be `secretToken`? Assuming it is meant to be the equivalent of https://www.elastic.co/guide/en/apm/agent/java/current/config-reporter.html#config-secret-token
I don't know if this differs for mobile/RUM, however.

LikeTheSalad:
Hi @trentm - Thanks for pointing it out! I was a bit confused because I saw "elastic server token" as environment variable [here](https://www.elastic.co/guide/en/apm/guide/current/open-telemetry.html#connect-open-telemetry-collector), and it made sense at the moment, but then I just took a deeper look and it's the only place with that name, so it seems like maybe it's mentioned in the docs like that by mistake 😅 - I've just updated this issue: #41 to apply this change as part of it, cheers!

trentm:
Ah, cool. Thanks. I'll ask apm-server folks if that `...SERVER_TOKEN` in https://www.elastic.co/guide/en/apm/guide/current/open-telemetry.html#connect-open-telemetry-collector should be updated.

PR #36, opened 2022-10-20
Add session attribute

LikeTheSalad:
Related to #27 

This PR creates an interface named `SessionIdProvider` which can be provided in the agent configuration in case our customers want to define their own logic on when to generate new IDs, and there's a `DefaultSessionIdProvider` which has a 30min timeout to generate new session ids as it's currently done in iOS. The default implementation is used if our customers don't provide their own implementation of `SessionIdProvider`.


android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/session/impl/DefaultSessionIdProvider.java
```diff
@@ -0,0 +1,67 @@
...
+        this(new SystemCurrentTimeMillisProvider());
+    }
+
+    @Override
+    public String getSessionId() {
+        verifySessionExpiration();
+        if (sessionId == null) {
+            sessionId = generateSessionId();
+        }
+        scheduleExpireTime();
```
felixbarny, line 50: At first, I was a little confused why we re-schedule on every access but I get it now - we want to generate a new ID after 30min of inactivity. Makes sense.

android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/session/impl/DefaultSessionIdProvider.java
```diff
@@ -0,0 +1,67 @@
...
+    DefaultSessionIdProvider(CurrentTimeMillisProvider currentTimeMillisProvider) {
+        this.currentTimeMillisProvider = currentTimeMillisProvider;
+    }
+
+    public DefaultSessionIdProvider() {
+        this(new SystemCurrentTimeMillisProvider());
+    }
+
+    @Override
+    public String getSessionId() {
```
felixbarny, line 45: This method may be called concurrently but it's not thread-safe.

android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/otel/processor/ElasticSpanProcessor.java
```diff
@@ -30,23 +32,27 @@
...
     }
 
     public ElasticSpanProcessor(SpanProcessor original) {
         this.original = original;
+        sessionIdProvider = ElasticApmAgent.get().configuration.sessionIdProvider;
     }
 
     @Override
     public void onStart(Context parentContext, ReadWriteSpan span) {
+        span.setAttribute(SESSION_ID_ATTRIBUTE_KEY, sessionIdProvider.getSessionId());
```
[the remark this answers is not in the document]
LikeTheSalad: That's a good point, I just checked and it seems like it's not thread safe according to this: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onstart - So I'll update it in here to avoid thread issues

android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/session/impl/DefaultSessionIdProvider.java
LikeTheSalad: Yes it's correct, this default behaviour is about having 30 mins as an idle timeout

android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/otel/processor/ElasticSpanProcessor.java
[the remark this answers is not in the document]
LikeTheSalad: I've just pushed some updates for it

felixbarny (approved):

Issue #37, opened 2022-10-20
Support Android Gradle plugin assets generation API changes


LikeTheSalad:
The Android gradle plugin API is currently going through big changes, one of them is related to the generation of assets files, needed to provide compile time information at runtime, such as service name, version and server parameters. The API for assets generation has changed in the latest version of the AGP (7.3.0), which makes our plugin to fail at compile time in projects that are using said version.

Most likely these kinds of issues will keep coming in future AGP releases, so ideally we should provide a flexible mechanism that would allow us to gracefully deal with them without having to discard host projects using older versions of the AGP.

For reference, this is the way it was done on version 7.2: https://github.com/android/gradle-recipes/blob/agp-7.2/Kotlin/addCustomAsset/app/build.gradle.kts

And this is the new API for version 7.3: https://github.com/android/gradle-recipes/blob/agp-7.3/Kotlin/addCustomAsset/app/build.gradle.kts

PR #43, opened 2022-10-28
Agp api changes compatibility

LikeTheSalad:
Related to #37 

This is a proposed approach to handle changes to the [Android Gradle Plugin](https://developer.android.com/studio/releases/gradle-plugin) API.

We're supporting it from its version `7.2.0`, and the latest published version is now `7.3.0`. Ever since they released their major version 7, they've been making breaking changes to their old API this year, and some of those are breaking the usage of the legacy API. Therefore we are using their new APIs that, unfortunately are still in development, and therefore are subject to changes.

One change to their Assets generation API happened in the new version `7.3.0`, currently used in our plugin to generate a properties file at compile time with relevant APM information such as the service name, version and server parameters. The outcome is the same, but some types used in version `7.2.0` have been moved to other locations, causing a `NoClassDefFoundError` exception when running our plugin on a project using AGP version `7.3.0`. Likewise, if we compile our plugin with the newest version, then a project using version `7.2.+` will throw the same exception.

So this PR aims to solve this issue by essentially providing an API of our own, flexible enough to accommodate to any API changes from the AGP. Our API, provided in the module `agp-compatibility-api`, has an interface `AgpCompatibilityEntrypoint` that receives the current AGP version of the host project for our plugin, and checks if said entry point is compatible with the running AGP version. If it is, it provides its `AgpCompatibilityManager` instance which contains use cases for each of the API changes that might happen in the future, Right now it has only one use case named `ApmInfoUseCase`.

An `AgpCompatibilityEntrypoint` implementation will be created for each AGP version change that might have breaking changes. Said implementation will lie in its own module where it will use its target AGP version to adapt the use cases to the new API. So right now we have 2 implementations, `Agp72CompatibilityEntrypoint` and `Agp73CompatibilityEntrypoint`, each in its own module (`agp-compatibility-7-2` and `agp-compatibility-7-3` respectively). `AgpCompatibilityEntrypoint` implementations are provided to our plugin as Java Services that will get loaded by the [ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) at runtime.

That's the proposed approach here in summary.

Some alternatives might be:
- Using reflection to use either API, though it comes with risks of difficult maintainability, scalability and readability.
- Writing these AGP API usage in Groovy and make use of its dynamic type features. But means that we would have several languages in our project, which might affect maintainability, plus Gradle [recommends to use statically typed languages](https://docs.gradle.org/current/userguide/designing_gradle_plugins.html#prefer_using_a_statically_typed_language_to_implement_a_plugin) for plugins.
- Drop lower AGP versions - This option isn't ideal, even though Google makes sure to force developers to keep their projects updated with the latest Android tools versions, version 7.2.0 is the second to last available one so it might be too much to drop that one and, considering that there are still objects within the new AGP API annotated with `@Incubating` it might mean that we'd probably have to keep on dropping versions as newer ones come in later if we follow this route.


felixbarny:
I haven't looked at the implementation but the approach sounds good to me.

felixbarny:
@LikeTheSalad did you intentionally close this PR instead of merging it?

LikeTheSalad:
Oh! no I didn't mean to close it, though it seems like it got closed after I deleted its target branch (I needed changes from that other branch in order to work on this one, so I chained them) but now that the target branch was merged into main, I can just change it here and reopen it to review it later. Thanks for the heads up!

PR #46, opened 2022-11-03
Adding documentation

### LINKED ISSUE #25: Add documentation

LikeTheSalad:
Closes #25


docs/faq.asciidoc
```diff
@@ -0,0 +1,44 @@
+[[faq]]
+== Frequently Asked Questions
+
+[float]
+[[faq-how-does-it-work]]
```
bryce-b, line 5: I think a lot of these need the `[descrete]` tag

bmorelli25:
Thanks, @LikeTheSalad! I'm going to push syntax fixes directly to your PR if that's okay.

docs/faq.asciidoc
bmorelli25: There are some subtle differences, but for the majority of our use cases `[descrete]` == `[float]`

bryce-b: Thanks @bmorelli25

LikeTheSalad:
No problem at all @bmorelli25! please push as many changes as you need :) and please let me know if there's anything I could help with as well!

docs/setup.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        ElasticApmAgent.initialize(this); <1>
+    }
+}
+----
+<1> Initialize the Elastic APM agent once.
+
```
bmorelli25, line 65: So step one is to add the plugin declaration to Gradle. Step two is initialize the agent in the Application class. What's step three? Does it just work™️? Is an application restart required?

docs/supported-technologies.asciidoc
```diff
@@ -0,0 +1,74 @@
...
+* <<supported-networking-frameworks>>
+
+[float]
+[[supported-agp-versions]]
+=== Android Gradle Plugin versions
+
+|===
+|Supported versions
+
+|7.2+, 7.3+
```
bmorelli25, line 19: Is this redundant? Or is it important to list both 7.2 and 7.3?

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+----
+// Android app's build.gradle file
+plugins {
+    //...
+    id "co.elastic.apm.android" version "[latest_version]"
+}
+
+elasticApm {
+    // Minimal configuration
+    serverUrl = "https://your.elastic.server"
```
bmorelli25, line 19: We explain the default for every config other than this one. What is the default value?

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+|secretToken
+|`ELASTIC_APM_SECRET_TOKEN`
+
+|===
+
+[float]
+=== Runtime configuration
+
+The runtime configuration is provided within your https://developer.android.com/reference/android/app/Application[Application] class at the time of initializing the Elastic agent.
+This configuration can override any previously-set compile time configuration.
```
bmorelli25, line 54: This _can_ override compile-time config? Or it _does_ overwrite compile-time config when set?

docs/faq.asciidoc
```diff
@@ -0,0 +1,44 @@
...
+Unlike other bytecode instrumentation libraries, Byte Buddy is designed so that it is impossible to corrupt the bytecode of instrumented classes.
+It also respects other transformations applied to your application at the same time.
+
+[float]
+[[faq-unsupported-technologies]]
+=== What if the agent doesn't support the technologies I'm using?
+
+The Agent is built on top of https://opentelemetry.io/docs/instrumentation/java/getting-started/[Open Telemetry] You can use its public API to create custom spans and transactions.
+
+To vote for prioritizing adding support for the technologies you are using, or TBD[get involved in the agent development]
```
bmorelli25, line 43: Are you planning on adding a google form or something here?

bmorelli25:
This is looking great! I've pushed some changes, but have a few questions before I keep going.

docs/setup.asciidoc
LikeTheSalad: That's a great point, I've been doing this for so long that I didn't think of how the overall init process is perceived the first time 😅
So, if you want to go with the default behaviour (which is to automatically instrument http requests and UI rendering as shown on the "supported technologies"), then yes, this is all you need to do because, step 3 might be to just compile (since it's an app, I think there's no "restart" concept, but instead a recompile or just compile process), which is something users will do anyway, but now that you mention it, I think it's better to clarify what they'll get out of those 2 steps and link them to the config in case they'd like to do some optional steps

docs/configuration.asciidoc
LikeTheSalad: Nice catch, it _does_ overwrite it, compile config stands no chance against the runtime one 😄

docs/faq.asciidoc
LikeTheSalad: I was, but I think for now I'll remove this and keep it simple because we're still trying to figure out the base functionality of the agent so far, so I don't think we're at the point where we ran out of ideas and have time to address more specific use cases 😅

docs/supported-technologies.asciidoc
LikeTheSalad: Well, though I'm not sure if 7.2+ would be perceived as only the patch versions for 7.2, or if it also would cover further minor versions too (7.3, 7.4... etc). In general we're just planning to support every AGP version starting from 7.2.0 onwards (even major versions too), I'm fine to change this to a broader term that doesn't require us to list all the major/minor versions, I just don't know what could be the best approach though, maybe something like `>= 7.2.0`?

LikeTheSalad:
Thanks @bmorelli25 - I've just pushed some changes based on your feedback, and also added some responses as well. Please let me know if any other questions or just feedback in general!

docs/supported-technologies.asciidoc
bmorelli25: I like `>= 7.2.0`. That's the syntax we use in the other agents.

docs/faq.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+
+[float]
+[[faq-how-does-it-work]]
+=== How does the agent work?
+
+The agent auto-instruments known frameworks and libraries and records interesting events, like HTTP requests.
+To do this, it leverages the capability of the JVM to instrument the bytecode of classes.
+This means that for the supported technologies, there are no code changes required.
+
+The agent automatically safely injects small pieces of code before and after interesting events to measure their duration and metadata like HTTP related information (like the URL, parameters, and headers).
```
bmorelli25, line 12:
```suggestion
The agent automatically and safely injects small pieces of code before and after interesting events to measure their duration and metadata, like HTTP-related information, including the URL, parameters, and headers.
```

docs/faq.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+[[faq-bytecode-instrumentation]]
+=== Is the agent doing bytecode instrumentation?
+
+Yes
+
+[float]
+[[faq-bytecode-instrumentation-safety]]
+=== How safe is bytecode instrumentation?
+
+Elastic APM is using the popular bytecode instrumentation library http://bytebuddy.net:[Byte Buddy], which takes care of the heavy lifting of dealing with actual bytecode and lets us write the instrumentation in pure Java.
```
bmorelli25, line 29:
```suggestion
Elastic APM uses the popular bytecode instrumentation library http://bytebuddy.net:[Byte Buddy], which takes care of the heavy lifting of dealing with actual bytecode and lets us write the instrumentation in pure Java.
```

docs/faq.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+It is created by a Java Champion, awarded with the Dukes Choice award and currently downloaded over 75 million times a year.
+
+Unlike other bytecode instrumentation libraries, Byte Buddy is designed so that it is impossible to corrupt the bytecode of instrumented classes.
+It also respects other transformations applied to your application at the same time.
+
+[float]
+[[faq-unsupported-technologies]]
+=== What if the agent doesn't support the technologies I'm using?
+
+The Agent is built on top of https://opentelemetry.io/docs/instrumentation/java/getting-started/[Open Telemetry] You can use its public API to create custom spans and transactions.
```
bmorelli25, line 41:
```suggestion
The agent is built on top of https://opentelemetry.io/docs/instrumentation/java/getting-started/[Open Telemetry] You can use its public API to create custom spans and transactions.
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+|secretToken
+|`ELASTIC_APM_SECRET_TOKEN`
+
+|===
+
+[float]
+=== Runtime configuration
+
+The runtime configuration is provided within your https://developer.android.com/reference/android/app/Application[Application] class at the time of initializing the Elastic agent.
```
bmorelli25, line 53:
```suggestion
The runtime configuration is provided within your https://developer.android.com/reference/android/app/Application[Application] class when initializing the Elastic agent.
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+This configuration overrides any previously-set compile time configuration.
+
+The way it works is by providing your own instance of the `ElasticApmConfiguration` class as shown below
```
bmorelli25, line 56:
```suggestion
Runtime configuration works is by providing your own instance of the `ElasticApmConfiguration` class as shown below:
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        ElasticApmAgent.initialize(this, ElasticApmConfiguration.builder().build());
+    }
+}
+----
+
+[float]
+==== APM server connectivity
```
bmorelli25, line 73:
```suggestion
==== APM Server connectivity
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+The server connectivity parameters can be provided at compile time, either by using the gradle dsl configuration, or by providing the server connectivity related environment variables as mentioned above.
+Later on, when the app is running, the connectivity parameters can be overriden by providing a custom `Connectivity` instance when initializing the Elastic agent.
```
bmorelli25, line 76:
```suggestion
The APM Server connectivity parameters can be provided at compile time, either by using the Gradle DSL configuration or by providing the APM Server connectivity-related environment variables as mentioned above.
Later on, when the app is running, the connectivity parameters can be overridden by providing a custom `Connectivity` instance when initializing the Elastic agent.
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+The `Connectivity` interface provides a set of convenient methods to avoid having to create your own implementation in case you need to define a custom connectivity.
```
bmorelli25, line 78: I'm not sure I understand what this sentence is trying to say.

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+        ElasticApmAgent.initialize(this, myCustomConnectivity);
+
+        // Optionally if you also define a custom configuration:
+        // ElasticApmAgent.initialize(this, ElasticApmConfiguration.builder().build(), myCustomConnectivity);
+    }
+}
+----
+
+[float]
+==== Application id configuration
```
bmorelli25, line 99: Should ID be capitalized here?

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+You can provide your application name and version dynamically by setting those when building your `ElasticApmConfiguration` instance as show below
```
bmorelli25, line 101:
```suggestion
You can provide your application name and version dynamically by setting those when building your `ElasticApmConfiguration` instance as show below:
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+
+The agent provides a configuration object for HTTP-related spans named `HttpTraceConfiguration`.
+You can pass an instance of it to the `ElasticApmConfiguration` object when initializing the agent in order to customize how the HTTP spans should be handled.
+
+[float]
+===== Filtering HTTP requests from getting traced
+
+By default, all of your app's HTTP requests will get traced.
+You can avoid some requests from getting traced by creating your own `HttpExclusionRule`.
+For example this is an exclusion rule that prevents all requests with the host `127.0.0.1` from getting traced:
```
bmorelli25, line 130:
```suggestion
For example, this is an exclusion rule that prevents all requests with the host `127.0.0.1` from getting traced:
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+class MyHttpExclusionRule extends HttpExclusionRule {
+
+    @Override
+    public boolean exclude(HttpRequest request) {
+        return request.url.getHost().equals("127.0.0.1");
+    }
+}
+----
+
+Then you'd need to add it to Elastic's Agent config through its `HttpTraceConfiguration` like so:
```
bmorelli25, line 143:
```suggestion
Then you'd need to add it to Elastic's Agent config through its `HttpTraceConfiguration`, like so:
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+----
+class MyHttpAttributesVisitor implements HttpAttributesVisitor {
+
+    public void visit(AttributesBuilder attrsBuilder, HttpRequest request) {
+        attrsBuilder.put("my_custom_attr_key", "my_custom_attr_value");
+    }
+}
+----
+
+Then you'd need to add it to Elastic's Agent config through its `HttpTraceConfiguration` like so:
```
bmorelli25, line 179:
```suggestion
Then you'd need to add it to Elastic's Agent config through its `HttpTraceConfiguration`, like so:
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+----
+
+[float]
+==== Trace spans attributes notes
+
+There are common attributes that the Elastic Agent gather for every Span, and those are listed https://github.com/elastic/apm/tree/main/specs/agents/mobile[here].
+
+However, due to the nature of Android's OS when it comes to collect some device-related data, some of the above mentioned resources require the Host app (your app) to have some runtime permissions granted.
+If their corresponding permissions aren't granted, then they won't be collected and nothing will be sent for those attributes.
+Those are the following:
```
bmorelli25, line 206: We have to be careful about saying "Elastic Agent" as that refers to the Elastic Agent, not the APM agent.

```suggestion
There are https://github.com/elastic/apm/tree/main/specs/agents/mobile[common attributes] that the Elastic APM agent gathers for every Span.
However, due to the nature of Android's OS, to collect some device-related data some of the above-mentioned resources require the Host app (your app) to have specific runtime permissions granted.
If the corresponding permissions aren't granted, then the device data won't be collected, and nothing will be sent for those attributes.
This table outlines the attributes and their corresponding permissions:
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,265 @@
...
+                .build();
+        ElasticApmAgent.initialize(this, configuration);
+    }
+}
+----
+
+[float]
+==== Advanced configurable options
+
+The configurable parameters provided by the Elastic agent aim to help configuring common use cases in an easy way, in most of the cases it means to act as a facade between your application and the Open Telemetry Java SDK that this agent is built on top.
```
bmorelli25, line 263:
```suggestion
The configurable parameters provided by the Elastic APM agent aim to help configuring common use cases in an easy way, in most of the cases it means to act as a facade between your application and the Open Telemetry Java SDK that this agent is built on top.
```

bmorelli25 (commented):
Thanks for addressing those questions. A few more suggestions for you 😃

docs/configuration.asciidoc
LikeTheSalad: Now that you mention it, it makes sense to capitalize it yeah! I just pushed the changes :)

docs/configuration.asciidoc
LikeTheSalad: I see, that's interesting. I think this might be the case of adding extra information that might cause confusion instead of helping much 😅 - So basically, `Connectivity` is an interface, so anyone can just create their own implementation and that would do. However, there are some common use cases for which I created helper methods, which might be easier to use rather than creating your own class that implements `Connectivity`. You can see those methods [here](https://github.com/elastic/apm-agent-android/pull/46/files#diff-6817141da84f46b7b26749d24c6bab4f40ace13bbc872a001df621335810613c).

I was thinking about maybe adding some examples here, but it really isn't necessary to use said helper methods to make it work, so I guess it'd be better if we just added a link to the `Connectivity` interface Javadoc, in case people are interested in seeing the code doc for it, where all the helper methods are described. It might be better as well for maintainability in the future, in case this interface changes, then this doc won't get inaccurate. However, for now I don't have the Javadoc link since the library isn't deployed yet, so for now I'm just going to remove it to avoid confusion 👍

LikeTheSalad:
Awesome cheers! I think I addressed them all, please let me know if there's any other feedback or questions!

PR #48, opened 2022-11-04
action: report test results

v1v:
Blocked by https://github.com/elastic/apm-pipeline-library/pull/1941

### What

Store the test results so they can be later on reported in the GitHub check.

### Why

Help to visualise the test reports

### Follow up

We will likely enable support for the OTEL reporting, as done in https://github.com/elastic/apm-pipeline-library/blob/main/.github/workflows/opentelemetry.yml so GitHub actions will represented as traces, in addition to also reporting the test results. But something to be done in the near future

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,11 @@ jobs:
         run: ./gradlew assemble
       - name: Test
         run: ./gradlew test && ./gradlew -p "android-test" test
+      - name: Debug
+        run: find . -name 'TEST-*.xml' -ls
       - name: Store test results
         if: success() || failure()
         uses: actions/upload-artifact@v3
         with:
           name: test-results
           path: "**/build/test-results/{test,testRelease*}/TEST-*.xml"
-
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,6 @@ jobs:
         run: ./gradlew assemble
       - name: Test
         run: ./gradlew test && ./gradlew -p "android-test" test
-      - name: Debug
-        run: find . -name 'TEST-*.xml' -ls
       - name: Store test results
         if: success() || failure()
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/test-reporter.yml
```diff
@@ -0,0 +1,20 @@
...
+    workflows:
+      - Continous Integration
+    types:
+      - completed
+
+jobs:
+  report:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dorny/test-reporter@c9b3d0e2bd2a4e96aaf424dbaa31c46b42318226
```
v1v, line 15:
```suggestion
      - uses: elastic/apm-pipeline-library/.github/actions/test-report@gh_actions
```

.github/workflows/ci.yml
```diff
@@ -20,4 +20,10 @@ jobs:
         run: ./gradlew assemble
       - name: Test
         run: ./gradlew test && ./gradlew -p "android-test" test
-        
+      - name: Store test results
+        if: success() || failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: test-results
+          path: "build/test-results/test/TEST-*.xml"
```
v1v, line 28: @LikeTheSalad , can you please help me to know where the JUnit test results can be found? I somehow thought it will be there, but it's not  :/

LikeTheSalad: Hi @v1v - looks good! although the `build` dir can be within other modules, so I think this should work: `**/build/test-results/{test,testRelease*}/TEST-*.xml` - Also, for Android projects, there's 2 build variants (debug and release), and both are tested using the same tests, so it's kind of a duplication of tests per variant, which is why I picked the `release` tests results only in the path I shared above. Feel free to modify it as you need :)

v1v:
```suggestion
          path: "**/build/test-results/{test,testRelease*}/TEST-*.xml"
```

v1v: Somehow its not found it yet

image

LikeTheSalad (approved):

PR #49, opened 2022-11-15
ci: script to install tools

v1v:
Ideally the release will need to run:

```bash
$ install-android-sdk.sh
$ export PATH=${PATH}:$PWD/.android-sdk/tools/bin/
$ export ANDROID_HOME=$PWD/.android-sdk
```

Then I ran `./gradlew test` locally and it worked like a charm

```
 '/Users/vmartinez/work/src/github.com/elastic/apm-agent-android/.android-sdk/tools')

BUILD SUCCESSFUL in 1m 14s
120 actionable tasks: 114 executed, 6 up-to-date
```

### DIFF
--- a/install-android-sdk.sh
+++ b/install-android-sdk.sh
@@ -24,11 +24,6 @@ rm ${COMMAND_LINE_TOOLS_ZIP}
 mv cmdline-tools latest
 ln -s ${SDKDIR}/cmdline-tools/latest ${SDKDIR}/tools
 
-echo Installing emulator...
-yes | ${ANDROID_HOME}/tools/bin/sdkmanager --install platform-tools emulator
-
 echo Installing platform SDK...
 yes | ${ANDROID_HOME}/tools/bin/sdkmanager --install "platforms;android-33"
 
-echo Starting ADB...
-${ANDROID_HOME}/platform-tools/adb devices
--- a/install-android-sdk.sh
+++ b/install-android-sdk.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+set -e
+
+SDKDIR=$PWD/.android-sdk
+export ANDROID_SDK_ROOT=${SDKDIR}
+export ANDROID_HOME=${SDKDIR}
+export ANDROID_AVD_HOME=${SDKDIR}/avd
+
+mkdir ${SDKDIR}
+mkdir ${SDKDIR}/cmdline-tools
+
+echo Downloading Android SDK...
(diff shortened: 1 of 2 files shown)


install-android-sdk.sh
```diff
@@ -0,0 +1,34 @@
...
+TYPE=linux
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  TYPE=mac
+fi
+curl -q https://dl.google.com/android/repository/commandlinetools-${TYPE}-8512546_latest.zip -o ${COMMAND_LINE_TOOLS_ZIP}
+unzip ${COMMAND_LINE_TOOLS_ZIP}
+rm ${COMMAND_LINE_TOOLS_ZIP}
+mv cmdline-tools latest
+ln -s ${SDKDIR}/cmdline-tools/latest ${SDKDIR}/tools
+
```
LikeTheSalad, line 26: I think at some point we'd need to accept the sdk's licenses? Maybe with this command:

```
yes | ${ANDROID_HOME}/tools/bin/sdkmanager --licenses
```

I remember it was a thing last time I had to do this, not sure if it's changed recently tho

install-android-sdk.sh
```diff
@@ -0,0 +1,34 @@
...
+echo Installing emulator...
+yes | ${ANDROID_HOME}/tools/bin/sdkmanager --install platform-tools emulator
```
LikeTheSalad, line 28: Probably we don't need to install the emulator

install-android-sdk.sh
```diff
@@ -0,0 +1,34 @@
...
+
+echo Installing platform SDK...
+yes | ${ANDROID_HOME}/tools/bin/sdkmanager --install "platforms;android-33"
+
+echo Starting ADB...
```
LikeTheSalad, line 33: I think we also don't need to start the ADB since we're not going to connect to android devices

install-android-sdk.sh
v1v, line 28:
```suggestion
```

install-android-sdk.sh
LikeTheSalad: After Victor's local test, it's confirmed that we don't need do add this line

install-android-sdk.sh
LikeTheSalad: I think we should leave the whole line except the `emulator` part at the end

LikeTheSalad (approved):

LikeTheSalad (approved):

PR #52, opened 2022-11-17
Add `page_header.html`

bmorelli25:
This PR adds a header to every documentation page explaining that this product is still under development. See the iOS docs for an example: https://www.elastic.co/guide/en/apm/agent/swift/current/index.html.

EDIT: I also added two small fixes to the docs.


LikeTheSalad (approved):
I'm really glad we do this because the very first release for sure is going to get massive breaking changes on further releases 😅 I hope people don't use it in prod

Issue #58, opened 2022-11-23
OTel desugaring


LikeTheSalad:
The Open Telemetry Java SDK mentions that for Android projects, the [desugaring config must be enabled](https://github.com/open-telemetry/opentelemetry-java#requirements).

Java language APIs/features apis are embedded in the Android OS, so when new ones are added into newer JDK versions (like what happened with version 8 where lambdas were added), a new Android OS version was needed to add support for them at runtime. However, some libraries (like OTel's), make use of "newer" JDK APIs, such as the [java.time.Duration in this case](https://github.com/open-telemetry/opentelemetry-java/blob/fa46f19d3936512b652a8a2c5d0be39f29066802/exporters/common/src/main/java/io/opentelemetry/exporter/internal/grpc/GrpcExporterBuilder.java#L139), which is available from the Android OS 8 (API level 26), so in devices with OS version < 8, there will be a crash of type "NoClassDefFoundError" when running them.

The way Google came up with a solution to avoid these issues, was to add an optional compilation process where new APIs would get "[desugared](https://developer.android.com/studio/write/java8-support#library-desugaring)" in order to make them compatible for older OS versions. They have a list of supported APIs for desugaring [here](https://developer.android.com/studio/write/java8-support-table).

So if our customers are targeting Android OS versions < 26, they must [opt in for desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring) in their app's `build.gradle` file.

According to Google's guide on how to add support for desugaring, it seems like it's a 2-step process, the first one is to enable a flag, and the second one is to add a dependency. However, the dependency's version to set cannot be the latest one all the time, since it depends on the version of the Android Gradle plugin that the project is on. So we must make sure to tell our customers to check on their project's AGP version before adding said library, or we could add all the "desugaring" configuration for them through our Gradle plugin, where we'd check our customer's AGP version and apply the best config for them automatically.

PR #59, opened 2022-11-23
58 otel desugaring

LikeTheSalad:
Related to #58 

Dropping support for Android API < 26 to help making the set up process easier. Might evaluate further options in the future that don't involve dropping < 26 support if needed.

### DIFF
--- a/android-test/app/build.gradle
+++ b/android-test/app/build.gradle
@@ -24,7 +24,7 @@ android {
     }
     defaultConfig {
         applicationId "co.elastic.apm.android.test"
-        minSdk 24
+        minSdk 26
         targetSdk 32
         versionCode 1
         versionName "1.0"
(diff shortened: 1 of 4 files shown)


felixbarny:
Is there a way to find out what the proportion of apps that have support Android API < 26 is?

felixbarny:
Am I reading this right that by dropping < 26, around 12% of all apps out there couldn't use the agent?

I think that's fine. At least for now, we shouldn't make our lives too hard in order to support older versions. I'd say let's start with only supporting API level 26+ and if we frequently get the feedback from users that need support for older versions, we can re-evaluate.

LikeTheSalad:
Kinda, I mean around 12% of the existing Android devices (acknowledged by Google) won't be able to run apps using our Agent... I'm not sure if that necessarily means 12% of the apps out there, though I think it should be a minority.

But yeah what you mentioned aligns with what I discussed earlier today with @AlexanderWert about it, we think it's better to do this to avoid the issues that supporting < 26 carries, because we're not sure if not supporting those older versions will actually be an issue for our customers. But if it is, and they let us know about it, then we can add that support, it would take a bit of work but at least we'd be sure that it's needed.

AlexanderWert (approved):

felixbarny (approved):

PR #60, opened 2022-11-23
Logging updates

LikeTheSalad:
- Filtering Android debug/verbose logs for android debug builds only
- Providing common logger factories id by default
- Providing function to create logger for class type


felixbarny:
Why do we have our own logging abstraction instead of using something like slf4j?

Issue #11, opened 2022-08-03
Publishing checklist


LikeTheSalad:
- [x] Use single Byte buddy version across all the subprojects.
- [x] Use release variant as maven publish module of the android-sdk subproject.
- [x] Check otel desugaring.
- [x] Check App Bundle files instrumentation.
- [x] Android assets generation compatible with all 7.2+ AGP versions
- [x] Make source headers generate on assembling
- [x] Update loggers default name
- [x] Update android logger level debug config
- [x] Running android-test from terminal
- [x] Adding tests for api breaking changes
- [x] Setting span type to "mobile"
- [x] Adding process.runtime attrs to resources
- [x] Adding okhttp spans tests
- [x] Adding "postDeploy" task

PR #61, opened 2022-11-23
Resources update

LikeTheSalad:
Related to #11 

- Adding missing resources `type`, `process.runtime.version` and `process.runtime.name`.
- Adding resources test coverage

### DIFF
--- a/android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/common/attributes/SdkIdVisitor.java
+++ b/android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/common/attributes/SdkIdVisitor.java
@@ -29,6 +29,7 @@ public class SdkIdVisitor implements AttributesBuilderVisitor {
     public void visit(AttributesBuilder builder) {
         builder.put(ResourceAttributes.TELEMETRY_SDK_NAME, "android")
                 .put(ResourceAttributes.TELEMETRY_SDK_VERSION, BuildConfig.APM_AGENT_VERSION)
+                .put("type", "mobile")
                 .put(ResourceAttributes.TELEMETRY_SDK_LANGUAGE, "java");
     }
 }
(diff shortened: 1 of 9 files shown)


felixbarny (approved):

android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/common/attributes/SdkIdVisitor.java
```diff
@@ -29,6 +29,7 @@ public class SdkIdVisitor implements AttributesBuilderVisitor {
     public void visit(AttributesBuilder builder) {
         builder.put(ResourceAttributes.TELEMETRY_SDK_NAME, "android")
                 .put(ResourceAttributes.TELEMETRY_SDK_VERSION, BuildConfig.APM_AGENT_VERSION)
+                .put("type", "mobile")
```
AlexanderWert, line 32: When is this `visit(...)`invoked? Once to setup the resource attributes or for each span?

I'm asking because, `TELEMETRY_SDK_NAME`, `TELEMETRY_SDK_VERSION` and `TELEMETRY_SDK_LANGUAGE` are resource attributes (so they have a device / process-wide scope). While `type -> mobile` should be a span attribute.

So, if this visit(...) method is used to build resource attributes, then setting the type `mobile` here is wrong. We should rather do it, when we start a UI root span.

AlexanderWert (changes requested):

android-sdk/src/main/java/co/elastic/apm/android/sdk/traces/common/attributes/SdkIdVisitor.java
LikeTheSalad: Oh I see, yes it's currently as you say, being set as a resource attr. I'll make the changes

AlexanderWert (approved):

PR #67, opened 2022-12-07
66 using ntp for internal clock

### LINKED ISSUE #66: Using NTP for internal clock
Through https://github.com/instacart/truetime-android

LikeTheSalad:
Closes #66 

This is to use [TrueTime](https://github.com/instacart/truetime-android) to get the real current time, if the NTP time isn't ready by the time a new span is created, then the device's system time will be used.

The NTP info will get cached until the device is rebooted, in which case the cache will get cleared and a new NTP info will get fetched next time the app gets open. (this behavior comes from TrueTime originally and has been replicated in our SDK)

TrueTime is not deployed into Maven Central but in JitPack instead, which causes an issue for our customers when adding our SDK into their projects, as our SDK only requires the Maven Central repo to be available (and Gradle Plugin portal for our plugin), but no JitPack is required, therefore the only ways they'd have to make our SDK work would be by:
- Adding JitPack to their own app's repos.
- Or, us packaging TrueTime within our own SDK.

We're packaging TrueTime into our SDK to avoid causing issues to our clients, however I'm open to alternatives in case the selected approach comes with issues that I might've missed. TrueTime classes aren't many, so the amount of extra weight it adds into our SDK is around 12KB.


android-sdk/src/main/java/co/elastic/apm/android/sdk/internal/otel/ElasticClock.java
```diff
@@ -0,0 +1,58 @@
...
+            Elog.getLogger().debug("Returning true time");
+            return TimeUnit.MILLISECONDS.toNanos(trueTimeWrapper.now().getTime());
+        }
+        Elog.getLogger().debug("Returning system time");
+        return TimeUnit.MILLISECONDS.toNanos(systemTimeProvider.getCurrentTimeMillis());
+    }
+
+    @Override
+    public long nanoTime() {
+        return systemTimeProvider.getNanoTime();
```
bryce-b, line 56: should this also be pulling from `trueTimeWrapper`?

LikeTheSalad: TrueTime doesn't seem to come with a nano time option, and it also seems like the [Clock.nanoTime](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/common/Clock.java#L40) function is only for "`time measurement with nanosecond precision that can only be used to calculate elapsed time.`" - So I think it's fine if it doesn't have the actual current time because the `now` function is the one used for that purpose, whereas `nanoTime` seems to be more for "stopwatch-like" purposes.

android-sdk/src/main/java/co/elastic/apm/android/sdk/internal/otel/ElasticClock.java
```diff
@@ -0,0 +1,58 @@
...
+        this.systemTimeProvider = systemTimeProvider;
+    }
+
+    public ElasticClock(TrueTimeWrapper trueTimeWrapper) {
+        this(trueTimeWrapper, SystemTimeProvider.get());
+    }
+
+    @Override
+    public long now() {
+        if (trueTimeWrapper.isInitialized()) {
```
AlexanderWert, line 46: can we end up with negative durations if this condition changes in the middle of a span? Or is the nanoTime() function always used for the span duration? Just wondering.

felixbarny: That's my understanding, too.

LikeTheSalad: I've confirmed that they use nanoTime to calculate the span duration when the [span is ended](https://github.com/open-telemetry/opentelemetry-java/blob/84170709c4d9ee0bb4c923e2fda41fd107645c4a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpan.java#L425) (they call [AnchoredClock.now()](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AnchoredClock.java#L50) which in turn calls Clock.nanoTime). I've also covered that case with [this test](https://github.com/elastic/apm-agent-android/pull/67/files#diff-d67c60a8ebb8521a0586bae0e5f653837491ba7ff209fc99803a938ac77286c9R49) just to make sure it stays the same in further OTel releases.

AlexanderWert (approved):

PR #68, opened 2022-12-12
Support daily snapshots

amannocci:
Signed-off-by: Adrien Mannocci <adrien.mannocci@elastic.co>

## What does this pull request do?

* Add support for daily snapshots.

### DIFF
--- a/.ci/snapshoty.yml
+++ b/.ci/snapshoty.yml
@@ -1,0 +1,64 @@
+---
+
+# Version of configuration to use
+version: '1.0'
+
+# You can define a Google Cloud Account to use
+account:
+  # Project id of the service account
+  project: '${GCS_PROJECT}'
+  # Private key id of the service account
+  private_key_id: '${GCS_PRIVATE_KEY_ID}'
+  # Private key of the service account
+  private_key: '${GCS_PRIVATE_KEY}'
+  # Email of the service account
+  client_email: '${GCS_CLIENT_EMAIL}'
+  # URI token
+  token_uri: 'https://oauth2.googleapis.com/token'
+
+x-metadata: &metadata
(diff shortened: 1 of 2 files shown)


reakaleek (approved):

.ci/snapshoty.yml
```diff
@@ -0,0 +1,34 @@
...
+  private_key: '${GCS_PRIVATE_KEY}'
+  # Email of the service account
+  client_email: '${GCS_CLIENT_EMAIL}'
+  # URI token
+  token_uri: 'https://oauth2.googleapis.com/token'
+
+# List of artifacts
+artifacts:
+  # Path to use for artifacts discovery 
+  - path: './android-plugin/build/libs'
```
LikeTheSalad, line 22: Hi! many thanks for this PR. Just one question, it seems like this would only upload the plugin jar file? I'm asking because, for the whole agent to work, all of the modules' binaries are going to be needed (`android-plugin`, `android-common`, `android-instrumentation`, `android-sdk`, `android-sdk-ktx` and all the submodules in `agp-compatibility`). So if the use case for these snapshots is to use them locally, I think we'll have to upload them all, because part of the `android-plugin`'s work is adding those other modules as the host project dependencies

cachedout (approved):

.ci/snapshoty.yml
amannocci: I updated the configuration to also push :
* android-common
* android-instrumentation
* android-plugin
* android-sdk
* android-sdk-ktx
* agp-compatibility

LikeTheSalad: Awesome, cheers! 
LGTM

LikeTheSalad (approved):

PR #90, opened 2023-01-18
action: more jobs in parallel and use caching

v1v:
https://github.com/marketplace/actions/gradle-build-action#caching

~ 7 minutes average https://github.com/elastic/apm-agent-android/actions/workflows/ci.yml

So let's see how it goes with this parallelisation


v1v:
I cannot see much advantage for running in parallel since the `test` takes 6 minutes while it takes 4 minutes with the previous approach.

PR #98, opened 2023-01-23
Update documentation

LikeTheSalad:


LikeTheSalad:
This is the first update to the docs since they were created, @bmorelli25 please take a look when you get a chance to make sure that the changes meet our quality standards

docs/configuration.asciidoc
```diff
@@ -206,20 +207,20 @@ This table outlines the attributes and their corresponding permissions:
...
 
-The Elastic agent provides a "session ID" for every span it creates.
+The Elastic agent provides a "session ID" for every span and log event it creates.
 The idea of a session is to provide a context that covers many transactions that a user did in order to fulfil their needs using an application.
-For most apps, a session could start when the user opens the app, and end when the user closes the app, or when the app is forced to get closed due to an unexpected error.
+For most apps, a session could start when the user opens the app, and end when the user closes the app.
 But for other apps, such as a ticketing app for a queue in a bank for example, the app will always be open, but a session might start when a person starts the process to get a new ticket, and end when the ticket is printed.
 
-By default, the Elastic agent generates a session ID on every app launch and keeps it until the app is terminated, or generates a new one after an idle period of time of 30 minutes.
+By default, the Elastic agent generates a session ID on app launch and keeps it until after an idle period of time of 30 minutes, restarting the timeout time on every APM-related transaction.
```
bmorelli25, line 223: This sentence was tough for me to digest. What about something like this?

```suggestion
By default, the APM agent generates a session ID on app launch. Each session is kept active until 30 minutes of idle time has elapsed. The idle time clock resets on every APM-related transaction.
```

docs/setup.asciidoc
```diff
@@ -69,15 +70,18 @@ class MyApp extends android.app.Application {
...
 [[compile-and-run]]
 === Compile and run
 
-All that's left is to compile and run your application. That's it!
+All that's left is to compile and run your application.
+That's it!
 
-After initializing the agent as shown above, you'll automatically create traces for all OkHttp network requests and for all https://developer.android.com/reference/android/app/Activity[Activity] and https://developer.android.com/reference/androidx/fragment/app/Fragment[Fragment] starting methods.
+After initializing the agent as shown above, you'll automatically create traces for all OkHttp network requests (including those created by tools that make use of OkHttp, such as Retrofit) and for all https://developer.android.com/reference/android/app/Activity[Activity] and https://developer.android.com/reference/androidx/fragment/app/Fragment[Fragment] starting methods.
+You'll also get crash reports sent automatically for when an unexpected error occurs in your app, and these events, as well as the traces already mentioned, will contain a "Session ID" that links the related events together in order to quickly get a clear view of any issue and its root cause.
```
bmorelli25, line 84: What if we move this content and the APM app/config/manual links into a new section called something like, "What's next?"? Here's what I'm thinking:

```suggestion
[float]
[[whats-next]]
=== What's next?

After initializing the agent as shown above, your application will automatically create traces for all OkHttp network requests (including those created by tools that make use of OkHttp, like Retrofit) and all https://developer.android.com/reference/android/app/Activity[Activity] and https://developer.android.com/reference/androidx/fragment/app/Fragment[Fragment] starting methods.
In addition, you'll get automatic crash reports when an unexpected error occurs in your app.

All of these events will contain a "Session ID" that links related events together—allowing you to make sense of and diagnose any issues that arise. Head to the **APM app in {kib}** to start exploring your data.

If you need to customize the Elastic APM agent to your project's needs, see <<configuration,configuration>>. If you need to create your own custom transactions, see <<manual-instrumentation, manual instrumentation>>.
```

docs/setup.asciidoc
```diff
@@ -69,15 +70,18 @@ class MyApp extends android.app.Application {
...
 
-Head over to the **APM app in {kib}** to see your Android application events,
-or see <<configuration,configuration>> for details on how to tailor the Elastic APM agent to your project's needs.
+Head over to the **APM app in {kib}** to see your Android application events, or see <<configuration,configuration>> for details on how to tailor the Elastic APM agent to your project's needs.
+Optionally, you can take a look at <<manual-instrumentation, manual instrumentation>> in case you need to create your own transactions as well.
```
bmorelli25, line 87: If you accept my above comment, then this content would move.

```suggestion
```

bmorelli25, line 87: If you accept my above comment, then this content would move.

```suggestion
```

bmorelli25, line 87: If you accept my above comment, then this content will move.

```suggestion
```

bmorelli25 (approved):
Looks good! I have a few recommendations.

docs/configuration.asciidoc
LikeTheSalad: Love it, thanks!

docs/setup.asciidoc
LikeTheSalad: I think it's great because as it is now, it seems like the setup process was cut off or something's missing, so the "What's next" title should help show where the setup ends and what they can do if that's not enough. Thanks!

PR #89, opened 2023-01-18
action: use snapshoty simple

v1v:
### What

Simplify the daily snapshots publishing by using the composite action.

It runs on push to `main`
Secrets are not available for PRs coming from forked repos.

### Why

Delegate the implementation details to the composite action.
Avoid errors when running PRs from forked repo

### Follow ups

Support for PRs but for now we only care about snapshots from the main branch.
Change branch to current in the action version

### Issues
Depends on https://github.com/elastic/apm-pipeline-library/pull/2016

PR #100, opened 2023-01-24
action: generate artifacts for snapshoty

v1v:
Partially implemented in https://github.com/elastic/apm-agent-android/pull/89 but required to assemble the artifacts so they can be uploaded.

Rather than copying them from `.github/workflows/ci.yml` I decided to generate them again. That will avoid glueing different workflows for different events.


v1v:
run elasticsearch-ci/docs

PR #104, opened 2023-01-30
action: run releases partially with GH actions

v1v:
### What

Use GitHub actions to release this project. 

### Why

Jenkins will be decommissioned in the near future, hence there is a need to implement the release outside of Jenkins. BuildKite is the one in charge of running releases. Hence, this proposal will use BK and hide the implementations by facilitating the release process through GH actions.

### Issues

Requires https://github.com/elastic/apm-pipeline-library/pull/2033

### How to use it?

Go to https://github.com/elastic/apm-agent-android/actions/workflows/release.yml
Choose the parameters
Then
1) You will receive a notification in slack
image

2) Then you can see the build running in BK:

image

3) And get a slack notification with the release status

### Follows up

- Create BK definition within this project for running the release, so there is no need to use any other GitHub repository
- Export Buildkite builds to a private Google Bucket or similar place, so there is no need to access Buildkite for queering the logs. Or even use Elastic for digesting those logs in favour of avoiding context switching in the CI. To be revisited.

### DIFF
--- a/.ci/release.sh
+++ b/.ci/release.sh
@@ -88,14 +88,12 @@ export COMMON_GRADLE_SIGNING_PARAMS="-Psigning.secretKeyRingFile=$SECRING_FILE -
 export COMMON_GRADLE_CONFIG_PARAMS="-Prelease=true -Pversion_override=${version_override_specifier}"
 export COMMON_GRADLE_DEPLOY_PARAMS="$COMMON_GRADLE_SIGNING_PARAMS $COMMON_GRADLE_CONFIG_PARAMS"
 
-if [ ${target_specifier} == 'all' ] || [ ${target_specifier} == 'mavenCentral' ]
-then
+if [[ "$target_specifier" == "all" ||  "$target_specifier" == "mavenCentral" ]]; then
   echo "--- Release the binaries to Maven Central"
   echo "./gradlew publishElasticPublicationToSonatypeRepository closeAndReleaseSonatypeStagingRepository $COMMON_GRADLE_DEPLOY_PARAMS"
 fi
 
-if [ ${target_specifier} == 'all' ] || [ ${target_specifier} == 'pluginPortal' ]
-then
+if [[ "$target_specifier" == "all" ||  "$target_specifier" == "pluginPortal" ]]; then
   echo "--- Release the binaries to the Gradle Plugin portal"
   echo "./gradlew publishPlugins -Pgradle.publish.key=$PLUGIN_PORTAL_KEY -Pgradle.publish.secret=$PLUGIN_PORTAL_SECRET $COMMON_GRADLE_DEPLOY_PARAMS"
 fi
--- a/.ci/release.sh
+++ b/.ci/release.sh
@@ -67,6 +67,8 @@ export PATH=${PATH}:$PWD/.android-sdk/tools/bin/
 export ANDROID_HOME=$PWD/.android-sdk
 
 ## Stage 4. Run release
+## TODO: let's stop here so we can test things work nicely until here
+exit 0
 
 set +x
 # Setting up common deploy params in env var
--- a/.ci/release.sh
+++ b/.ci/release.sh
@@ -11,14 +11,16 @@ set -e
 
 ## Stage 0. Prepare vault context to access the secrets
 set +x
-export VAULT_ROLE_ID=$(vault read -field=role-id secret/ci/elastic-observability-robots-playground/internal-ci-approle)
-export VAULT_SECRET_ID=$(vault read -field=secret-id secret/ci/elastic-observability-robots-playground/internal-ci-approle)
-export VAULT_ADDR=$(vault read -field=vault-url secret/ci/elastic-observability-robots-playground/internal-ci-approle)
-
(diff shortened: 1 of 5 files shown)


v1v:
@LikeTheSalad and @elastic/observablt-ci , what are your thoughts? This PR will need further changes in terms of passing parameters to the BK GH action, but I'd like to gather your feedback, 🙏

LikeTheSalad:
Looks good to me! I like the idea of having the `release.sh` script in this repo since it could make things smoother when it comes to update the overall CD process

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+
+    steps:
+      - id: buildkite
+        name: Run Release
+        # TODO: for testing purposes let's use the feature branch rather than current
+        uses: elastic/apm-pipeline-library/.github/actions/buildkite@feature/avoid-guessing-git-commit
```
v1v, line 42:
```suggestion
        uses: elastic/apm-pipeline-library/.github/actions/buildkite@current
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+            target_specifier=${{ inputs.target_specifier || 'all' }}
+            version_override_specifier=${{ inputs.version_override_specifier || '' }}
+
+      - if: ${{ always() }}
+        uses: elastic/apm-pipeline-library/.github/actions/slack-message@current
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          channel: "#on-week-oblt-productivity"
```
v1v, line 62:
```suggestion
          channel: "#apm-agent-mobile"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    steps:
+      - id: buildkite
+        name: Run Release
+        uses: elastic/apm-pipeline-library/.github/actions/buildkite@current
+        with:
+          vaultUrl: ${{ secrets.VAULT_ADDR }}
+          vaultRoleId: ${{ secrets.VAULT_ROLE_ID }}
+          vaultSecretId: ${{ secrets.VAULT_SECRET_ID }}
+          pipeline: observability-release
```
v1v, line 44: Eventually this particular pipeline will be migrated to be defined within this repository

.ci/release.sh
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env bash
+## This script runs the release given the different environment variables
```
v1v, line 2: This particular script was initially defined in JJB. It has been adapted to support running on Buildkite, hence there are some specific echo and env variables in place to ensure the logs UI can be collapsed and no secrets are printed in the logs.

.ci/release.sh
```diff
@@ -0,0 +1,107 @@
...
+env | sort
+
+echo "--- Prepare vault context"
+set +x
+VAULT_ROLE_ID_SECRET=$(vault read -field=role-id secret/ci/elastic-observability-robots-playground/internal-ci-approle)
+export VAULT_ROLE_ID_SECRET
+VAULT_SECRET_ID_SECRET=$(vault read -field=secret-id secret/ci/elastic-observability-robots-playground/internal-ci-approle)
+export VAULT_SECRET_ID_SECRET
+VAULT_ADDR=$(vault read -field=vault-url secret/ci/elastic-observability-robots-playground/internal-ci-approle)
+export VAULT_ADDR
```
v1v, line 23: This is required to access the secrets in `vault-ci` as they cannot access the `vault-prod`.

Eventually the secret location will change or likely we will use a different approach for the releases as we are in discussion with the release team and they might change the way it's done as today.

LikeTheSalad (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+          buildEnvVars: |
+            repo=${{ github.repository }}
+            commit=${{ github.sha }}
+            command=.ci/release.sh
+            channel='#apm-agent-mobile'
+            branch_specifier=${{ inputs.branch_specifier || 'main' }}
+            target_specifier=${{ inputs.target_specifier || 'all' }}
+            version_override_specifier=${{ inputs.version_override_specifier || '' }}
+
+      - if: ${{ always() }}
```
reakaleek, line 56: is `always` the desired behaviour?

What if the buildkite action fails for any reason?

v1v: Buildkite pipeline will be the one notifying if things went wrong or fine. This slack will just notify that something has been triggered in BK regardless of the status. I could split in two different notifications, but somehow I though `buildkite action` will never fail. but it could if wrong parameters.

Let me put the logic in place to play safe

reakaleek: it could also fail if there was a network error or something, right?

v1v: See 161568e

reakaleek (approved):

PR #109, opened 2023-02-01
release: define buildkite pipeline

v1v:
Caused by https://github.com/elastic/apm-agent-android/pull/104 so we can use the specific release pipeline in Buildkite.

This will be merged only once we validate the first release has done correctly in BK using the playground.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,14 +41,10 @@ jobs:
           vaultUrl: ${{ secrets.VAULT_ADDR }}
           vaultRoleId: ${{ secrets.VAULT_ROLE_ID }}
           vaultSecretId: ${{ secrets.VAULT_SECRET_ID }}
-          pipeline: observability-release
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -41,14 +41,10 @@ jobs:
...
-          pipeline: observability-release
+          pipeline: apm-agent-android-release
           waitFor: false
           printBuildLogs: false
           buildEnvVars: |
-            repo=${{ github.repository }}
-            commit=${{ github.sha }}
-            command=.ci/release.sh
-            channel='#apm-agent-mobile'
             branch_specifier=${{ inputs.branch_specifier || 'main' }}
```
v1v, line 48: This variable should be honoured in `.ci/release.sh`

reakaleek (approved):

LikeTheSalad (approved):

PR #115, opened 2023-02-02
release dry run report logs to google bucket

v1v:
### What

Support dry-run executions so we can ensure all the moving pieces (access to secrets, gpg configuration, tools installation, git checkout) work as expected.

Additionally, I added the upload-artifact script to google (to be implemented in a follow up) so it can help us to see those logs somewhere else without the need to access to Buildkite.

### DIFF
--- a/.ci/release.sh
+++ b/.ci/release.sh
@@ -3,6 +3,7 @@
 ##  branch_specifier
 ##  target_specifier
 ##  version_override_specifier
+##  dry_run
 ##
 ##  NOTE: *_SECRET env variables are masked, hence if you'd like to avoid any
 ##        surprises please use the suffix _SECRET for those values that contain
@@ -23,6 +24,7 @@ VAULT_ADDR=$(vault read -field=vault-url secret/ci/elastic-observability-robots-
 export VAULT_ADDR
 
 # Delete the vault specific accessing the ci vault
+export VAULT_TOKEN_PREVIOUS=$VAULT_TOKEN
 unset VAULT_TOKEN
 
 echo "--- Prepare release context"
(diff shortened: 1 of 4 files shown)


reakaleek (approved):

.ci/release.sh
```diff
@@ -104,14 +107,26 @@ export COMMON_GRADLE_DEPLOY_PARAMS="$COMMON_GRADLE_SIGNING_PARAMS $COMMON_GRADLE
 
 if [[ "$target_specifier" == "all" ||  "$target_specifier" == "mavenCentral" ]]; then
   echo "--- Release the binaries to Maven Central"
-  ./gradlew publishElasticPublicationToSonatypeRepository closeAndReleaseSonatypeStagingRepository $COMMON_GRADLE_DEPLOY_PARAMS
+  if [[ "$dry_run" == "true" ]] ; then
+    echo './gradlew publishElasticPublicationToSonatypeRepository closeAndReleaseSonatypeStagingRepository'
```
LikeTheSalad, line 111: Just so you know (probably it's not an issue for the purposes of this "dry-run" feature) but without the signing key parameters this command will fail

v1v: I didn't want to echo any potential variable which might be interpolated with some sensitive data. I've just found that the only way to mask things with Buildkite is by using some pre-scripts that need to be stored somewhere else in the repository.

LikeTheSalad: Oh! sorry I didn't realize it's just printing it, not actually running the command 😅

LikeTheSalad (approved):

PR #117, opened 2023-02-02
release: upload artifacts

v1v:
`artifact_paths` seems to be the way to upload the artifacts though I need to understand whether the location is somehow predictable, so we can provide those details when the slack notification will happen

### DIFF
--- a/.buildkite/release.yml
+++ b/.buildkite/release.yml
@@ -5,10 +5,7 @@ steps:
   - label: "Run the release"
     key: "release"
     commands: .ci/release.sh | tee release.out
-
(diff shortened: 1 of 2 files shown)


.buildkite/release.yml
```diff
@@ -5,10 +5,7 @@ steps:
   - label: "Run the release"
     key: "release"
     commands: .ci/release.sh | tee release.out
```
v1v, line 7: https://buildkite.com/docs/pipelines/managing-log-output#storing-the-original-log

LikeTheSalad (approved):

PR #137, opened 2023-03-24
Updating docs

LikeTheSalad:


docs/configuration.asciidoc
```diff
@@ -218,3 +249,10 @@ This table outlines the attributes and their corresponding permissions:
 The configurable parameters provided by the Elastic APM agent aim to help configuring common use cases in an easy way, in most of the cases it means to act as a facade between your application and the OpenTelemetry Java SDK that this agent is built on top.
 If your project requires to configure more advanced aspects of the overall APM processes, you could directly apply that configuration using the https://opentelemetry.io/docs/instrumentation/java/getting-started/[OpenTelemetry SDK], which becomes available for you to use within your project by adding the Elastic agent plugin, as explained in <<setup,the agent setup guide>>.
 Said configuration will be used by the Elastic agent for the https://opentelemetry.io/docs/concepts/signals/[signals] it sends out of the box.
+
+[float]
+=== Central configuration
+
+Some of the runtime configurable options are available through Kibana's https://www.elastic.co/guide/en/kibana/master/agent-configuration.html[central configuration].
```
bmorelli25, line 256:
```suggestion
Some of the runtime configurable options are available through Kibana's {kibana-ref}/agent-configuration.html[central configuration].
```

docs/configuration.asciidoc
```diff
@@ -218,3 +249,10 @@ This table outlines the attributes and their corresponding permissions:
...
+
+You can take a look at the available options https://github.com/elastic/apm/blob/main/specs/agents/mobile/configuration.md[here].
```
bmorelli25, line 258: Why aren't we documenting this in the docs? With the other agents, we have a `dynamic` badge that we add to any configurations that can be changed dynamically (like with central config). Here's an example: https://www.elastic.co/guide/en/apm/agent/java/current/configuration.html.

LikeTheSalad: Sounds good! I didn't think much of it other than just reusing what's available elsewhere, to keep it as a single source of truth. Though now that I give it another thought, maybe that single source of truth, at least for these config options, should be here instead 🤔 it should be easier to read by avoiding jumping back and forth across different places. I'll make the changes 👍

bmorelli25: Awesome, thanks! Let me know if you need help with anything.

LikeTheSalad: I've just made the update, I _think_ it should be alright, although there's one line that reads `[options="header"]` that I'm not sure if it's needed. That's the only part I wasn't sure about when looking at Java Agent's docs.

docs/configuration.asciidoc
```diff
@@ -115,6 +116,36 @@ class MyApp extends android.app.Application {
...
+        ElasticApmConfiguration configuration = ElasticApmConfiguration.builder()
+                .setInstrumentationConfiguration(instrumentations)
+                .build();
+        ElasticApmAgent.initialize(this, configuration);
+    }
+}
+----
+
+Please note that when building an InstrumentationConfiguration object using `InstrumentationConfiguration.builder()`
+all the instrumentations are disabled by default, so you only need to enable the ones you need using the builder setter methods.
```
bmorelli25, line 147:
```suggestion
NOTE: When building an InstrumentationConfiguration object using `InstrumentationConfiguration.builder()`,
all instrumentations are disabled by default. Only enable the instrumentations you need using the builder setter methods.
```

docs/configuration.asciidoc
```diff
@@ -218,3 +249,34 @@ This table outlines the attributes and their corresponding permissions:
...
+[[configuration-dynamic]]
+=== Dynamic configuration image:./images/dynamic-config.svg[]
+
+Configuration options marked with Dynamic true can be changed at runtime when set from Kibana's {kibana-ref}/agent-configuration.html[central configuration].
+
+[float]
+=== Option reference
+
+This is a list of all configuration options.
+Click on a key to get more information.
```
bmorelli25, line 263: Is this supposed to be here?

docs/configuration.asciidoc
bmorelli25: Yup, looks good! The `[options="header"]` line bolds the top row of the table you created; I'd leave it there.

docs/configuration.asciidoc
LikeTheSalad: I'm not sure, seems like in the Java Agent docs the options become clickable. So it was mostly because it's like this over there, but based on what you mentioned here it seems it's not really needed so I'm fine with removing it.

bmorelli25 (approved):

bmorelli25:
Once version 0.4.0 is released, I can add a link to this page: https://www.elastic.co/guide/en/kibana/8.6/agent-configuration.html. If you can let me know in this issue after the release, I'd appreciate it 😄  https://github.com/elastic/observability-docs/issues/2812

PR #160, opened 2023-04-28
Add signal filtering api

### LINKED ISSUE #158: Add signal filtering API
Provide configurable filters for each signal. Each filter interface would provide a method named `shouldInclude` that contains the final version of the signal's data object (before it gets sent to the exporter). Said function should return a boolean, where `true` would mean that the element will get sent to the backend, and `false` to filter it out.

LikeTheSalad:
Closes #158

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -162,7 +162,7 @@ class MyApp extends android.app.Application {
 ----
 
 Each filter will contain a `shouldInclude` function which provides the signal item to be evaluated.
-This function must return a boolean value, `true` for when the provided item is allowed to continue, `false` when it must be discarded.
+This function must return a boolean value--`true` when the provided item is allowed to continue or `false` when it must be discarded.
 
 You can add multiple filters per signal which will be iterated over (in the order they were added) until all the filters are checked or until one of them decides to discard the signal item provided.
 


android-sdk/src/main/java/co/elastic/apm/android/sdk/ElasticApmConfiguration.java
```diff
@@ -122,6 +134,36 @@ public Builder setSignalConfiguration(SignalConfiguration signalConfiguration) {
             return this;
         }
 
+        /**
```
LikeTheSalad, line 137: In the end, I figured that, at least in Android, setting the filters with a function named `set*` might be better in order to keep consistency with all the other config options @bryce-b @gregkalapos jsyk

bryce-b:
I think we should allow for filter chaining. By this I mean allowing multiple filters to be set. Instead of `setFilter` we can have `addFilter`. I think this will allow customers a cleaner looking filter set up, if they want to have multiple checks.

LikeTheSalad:
Sounds good, I just added filter chaining support @bryce-b

bryce-b (approved):

docs/configuration.asciidoc
```diff
@@ -127,6 +127,45 @@ class MyApp extends android.app.Application {
 }
 ----
 
+[float]
+==== Signal filtering
+
+You can provide your own filters to decide what spans, logs and metrics are allowed to be exported to the backend.
```
bmorelli25, line 133:
```suggestion
You can provide your own filters to specify which spans, logs, and metrics are allowed to be exported to the backend.
```

docs/configuration.asciidoc
```diff
@@ -127,6 +127,45 @@ class MyApp extends android.app.Application {
...
+                    }
+                })
+                .build();
+        ElasticApmAgent.initialize(this, configuration);
+    }
+}
+----
+
+Each filter will contain a `shouldInclude` function which provides the signal item to be evaluated.
+This function must return a boolean value, `true` for when the provided item is allowed to continue, `false` when it must be discarded.
```
bmorelli25, line 165:
```suggestion
This function must return a boolean value--`true` when the provided item is allowed to continue or `false` when it must be discarded.
```

bmorelli25 (approved):

Issue #94, opened 2023-01-19
Temporarily store signals in disk


LikeTheSalad:
To send them later. This would help for the following use cases:

- Retrying failed attempts to contact our backend
- More reliable crash reporting
- Keeping track of the activity that happened while the device didn't have internet connectivity

PR #163, opened 2023-05-16
Serializing logs

LikeTheSalad:
Related to #94 

A lot of files have been generated from OTel proto files.


bryce-b:
Are these files mostly duplicated from the otel sdk? Do they need to be brought into our project because they are private in the otel sdk?

bryce-b (approved):

felixbarny:
I don't think duplicating these files is a great long-term solution.

As on-disk buffering seems to be a common use case, can we contribute that feature to the upstream OTel SDK?

LikeTheSalad:
I'm up for contributing, the feature you mean is the whole disk-buffering? or this serialization one? Maybe the serialization itself might be a good feature to provide, though I'm not sure if the whole disk buffering one would be too feasible to get into the main SDK, since there might be some specific needs depending on whether it is done on Android or not. I know there are some differences in regards to file handling in Android, APIs and permissions, that are exclusive to it. I haven't taken a deep look though, maybe there's no issue, or maybe a lot of disk-managing processes can be abstracted so that each platform can provide its implementation...

LikeTheSalad:
I just found [this repo](https://github.com/open-telemetry/opentelemetry-proto-java), which wasn't listed in the main [otel-java releases](https://github.com/open-telemetry/opentelemetry-java#releases), which apparently just automatically fetches the protos, creates the Java files and releases them to maven central. I'm going to check if we can reuse these, though it doesn't seem like these are the exact ones used in the main OTel SDK, because the [package name](https://github.com/open-telemetry/opentelemetry-java/blob/06cc7c1d67c4a02f69df3390eebdbf4ac0632462/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/otlp/ResourceMarshaler.java#L12) is different from the ones in the generated lib (they might be shadowing those) so in the end at runtime there would still be duplication, though at least it won't be in our project's source

LikeTheSalad:
Here are some interesting notes on how they decided to do serialization in the OTel Java SDK: https://github.com/open-telemetry/opentelemetry-specification/discussions/1996 - Apparently they aren't relying in the generated proto source to do it, It seems like the lib size was an important reason for that, especially for Android cases, which makes sense if using the full [protobuf library](https://github.com/protocolbuffers/protobuf/tree/main/java#maven), instead of the [lite version](https://github.com/protocolbuffers/protobuf/tree/main/java#use-java-protocol-buffers-on-android). I'm planning to join this week's Java SIG to see if I can gather more information on this.

felixbarny:
That would be better, although I really think we should try to contribute disk buffering upstream. We can either find a way that works for both Android and regular Java, or find a common abstraction and have two different implementations or even something more generic that lets you plug in different persistence mechanisms, such as an embedded H2 database. But if there's an easy way to just buffer via local files that work for both Android and Java, that would be the preferred option IMO.

LikeTheSalad:
I agree it would be very useful to have that in the main OTel SDK. Let me check in the Java SIG what's their stance on having it as an extra exporter or something the like.

LikeTheSalad:
This feature is intended to be contributed to OTel, so we're closing this PR.

PR #169, opened 2023-06-14
Updating cicd

LikeTheSalad:


.github/workflows/release.yml
```diff
@@ -39,6 +39,16 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: elastic/apm-pipeline-library/.github/actions/github-token@main
```
LikeTheSalad, line 42: I saw this being done [in here](https://github.com/elastic/apm-agent-java/pull/3178/files) and, since the use case is similar for this repo, I'd like to apply the same git auth process.

kuisathaverat, line 42: `current` points to the latest stable version

```suggestion
      - uses: elastic/apm-pipeline-library/.github/actions/github-token@current
```

.github/workflows/release.yml
```diff
@@ -39,6 +39,16 @@ jobs:
...
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+      - uses: elastic/apm-pipeline-library/.github/actions/setup-git@current
+        with:
+          username: ${{ env.GIT_USER }}
+          email: ${{ env.GIT_EMAIL }}
```
kuisathaverat, line 50: These parameters will be calculated by the action

```suggestion
```

kuisathaverat:
@elastic/ci-robots Could you review this one?

LikeTheSalad:
There are a lot of files in this PR that are specific to this repo's custom functionality, except for the `.github/workflows/release.yml` one, which is the one I wasn't too sure about since I'm not aware of the details of the apmmachine Git auth functionality. So since it seems like, at least from a general perspective, there are no apparent issues that you see with it @kuisathaverat, I'm inclined to move forward with merging these changes, then try them out in a new release and, if the Git auth fails, I think we can then create a PR that would be focused only on the `release.yml` file, and that way it might be more reviewer-friendly for anyone not aware of the specifics of this repo.

.github/workflows/release.yml
```diff
@@ -39,6 +39,16 @@ jobs:
...
+      - uses: elastic/apm-pipeline-library/.github/actions/github-token@current
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+      - uses: elastic/apm-pipeline-library/.github/actions/setup-git@current
+        with:
+          username: ${{ env.GIT_USER }}
+          email: ${{ env.GIT_EMAIL }}
+          token: ${{ env.GITHUB_TOKEN }}
```
reakaleek, line 51: As far as I can see, the added token has no effect on the workflow.

I guess you are trying to add a GitHub PAT. But your adding it in the workflow.
However, the this job delegates it to a Buildkite job.

reakaleek: This is the corresponding Buildkite pipeline https://github.com/elastic/apm-agent-android/blob/main/.buildkite/release.yml

amannocci: The `GITHUB_TOKEN` is the one set up by the `github-token` GitHub Actions.

LikeTheSalad: Got it, thanks @reakaleek  - I'll move this to the `.buildkite/release.yml` file then.

@amannocci  I'm not sure I'm following your comment, does that mean that the `setup-git` GitHub action can read the `GITHUB_TOKEN` without me having to pass it in line 51? Or that I'm not passing the token to the `setup-git` action correctly?

amannocci: Sorry, it was the explanation for the following comment:

> As far as I can see the added token has no effect on the workflow.
> 
> I guess you are trying to add a GitHub PAT. But your adding it in the GH workflow. However, this job delegates it to a Buildkite job.

In fact, we are using the `GITHUB_TOKEN` set up by the GitHub Action `github-token` at **L51**.
On another hand, if we use `${{ secrets.GITHUB_TOKEN }}` instead of `${{ env.GITHUB_TOKEN }}` then we will use the auto-generated Github Token set up by GitHub Action.

LikeTheSalad: I see thank you. And for my use case (which is to let the buildkite job to push changes straight into the `main` branch as `apmmachine`) is it preferable to use the token from the `github-token` action? Or is it better to use the auto-generated by Github one?

reakaleek: the generated token is not able to push to protected branches. See [discussion](https://github.com/orgs/community/discussions/25305).

Hence, you actually have to use the PAT of `apmmachine`

LikeTheSalad: Got it, thank you @reakaleek - I've just made the change to move this over to `.buildkite/release.yml`.

.buildkite/release.yml
```diff
@@ -2,6 +2,16 @@ agents:
...
+  - uses: elastic/apm-pipeline-library/.github/actions/github-token@current
+    with:
+      url: ${{ secrets.VAULT_ADDR }}
+      roleId: ${{ secrets.VAULT_ROLE_ID }}
+      secretId: ${{ secrets.VAULT_SECRET_ID }}
+  - uses: elastic/apm-pipeline-library/.github/actions/setup-git@current
+    with:
+      username: ${{ env.GIT_USER }}
+      email: ${{ env.GIT_EMAIL }}
+      token: ${{ env.GITHUB_TOKEN }}
```
reakaleek, line 14: Sorry if I created any confusion. 
However, this is GH workflow syntax. This won't work in a buildkite pipeline.

reakaleek: Here is example on how secrets are currently acquired from vault.

https://github.com/elastic/apm-agent-android/blob/b42538638bd521297bf7a9f9f855150fc48ae3c7/.ci/release.sh#L62

You would need to get the token in the same manner.

Unfortunately, I don't know if there is a token for apmmachine stored in the ci vault.

@elastic/observablt-ci anyone of you knows how to get the secret for `apmmachine` in buildkite?

LikeTheSalad: Oh! I see, tbh I thought it seemed too easy to be true 😅 - I've been mostly looking at how this has been done in the Java Agent repo. And after a closer look, it seems like the code change pushed into main is done outside of Buildkite, so I'm thinking that maybe that can be done here as well, in case getting this token inside Buildkite might not be worth the trouble... So I'll take a look at that alternative in the meantime, cheers.

amannocci: > @elastic/observablt-ci anyone of you knows how to get the secret for `apmmachine` in buildkite?

Buildkite can retrieve secrets from another vault.
I will send the documentation by DM since it's internal.

LikeTheSalad: I just moved the post-deploy command to the GH release.yml file [here](https://github.com/elastic/apm-agent-android/pull/169/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R69). Alongside the GIT auth set up as well. I think it should work. Any thoughts @reakaleek @amannocci ?

LikeTheSalad: Actually, thinking about it a bit further, I might have to pull the agent project in the Action before running that command.

LikeTheSalad: If I add another job in the GH release.yml file (there's currently only one, which is the `release` one) is it possible to configure that second job to run after the `release` one only if it succeeded?

LikeTheSalad: I've added a new GH action job that should run after the release has been successful. In this new job we should do the version bump and changelog update and it should all get pushed straight into the `main` branch. I think it should work the way I wrote it, if there are any issues please let me know @amannocci @reakaleek - Otherwise I'm planning to merge this by the end of this week and apply future adjustments if needed since this PR has been open for too long now.

reakaleek: I think you need to set `waitFor` to `true`

https://github.com/elastic/apm-agent-android/blob/947996594fa182ce11f5cb4d93d76430b75a657d/.github/workflows/release.yml#L50

so this works as expected

reakaleek: Another thing I noticed:

pushing to protected branches does not work out of the box in github actions.

You need to use a GH Personal Access Token that has elevated permissions.

reakaleek: You can find an example job here with the necessary setup to do pushing to a protected branch:


https://github.com/elastic/apm-agent-java/blob/main/.github/workflows/release.yml#L128-L151


Also, we need to add @apmmachine as a collaborating with a role that is able to bypass branch protection.

.github/workflows/release.yml
```diff
@@ -75,3 +75,25 @@ jobs:
...
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ env.TAG_NAME }}
```
reakaleek, line 97: I guess this is a copy paste error. I think this should be something different.

maybe `inputs.branch_specifier` or just hardcoded `main`?

reakaleek (changes requested):

.github/workflows/release.yml
LikeTheSalad: I didn't notice, thanks for the heads-up! It should be fine now.

reakaleek (approved):

kuisathaverat (approved):

PR #173, opened 2023-08-11
Adding network information to logs

### LINKED ISSUE #140: Add network information to logs

LikeTheSalad:
Closes #140 

This PR adds `net.host.connection.type` (and `net.host.connection.subtype` when available) to all log signals. These attr names arrive at ES as `network.connection.type` and `network.connection.subtype` respectively.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,8 @@ ${next_release_notes}
 ===== Features
 
 * Updating OpenTelemetry SDK to 1.27.0 where logs are stable: {pull}168[#168]
+* Updating OpenTelemetry SDK to 1.28.0 where the new disk buffering lib is present: {pull}170[#170]
+* Adding network connectivity attributes to logs: {pull}?[#?]
 ////
 
 [[release-notes-0.7.0]]
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -34,7 +34,7 @@ ${next_release_notes}
 
 * Updating OpenTelemetry SDK to 1.27.0 where logs are stable: {pull}168[#168]
 * Updating OpenTelemetry SDK to 1.28.0 where the new disk buffering lib is present: {pull}170[#170]
-* Adding network connectivity attributes to logs: {pull}?[#?]
+* Adding network connectivity attributes to logs: {pull}173[#173]
 ////
 
 [[release-notes-0.7.0]]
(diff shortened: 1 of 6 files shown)


felixbarny:
I was under the impression that we just wanted to add network information to errors and crashes, not all logs.

LikeTheSalad:
Yes, we did mention that last week, though yesterday we decided to do so for all logs. If I remember correctly I think it was due to the fact that our users can send custom errors by sending logs with a "type=error" attribute (or something similar, I can't remember the exact attr right now), so if we added these extra attrs only to the crashes and errors that we create, then the custom errors would have the same filtering issue.

LikeTheSalad:
The visitor gets called every time a log is [emitted](https://github.com/open-telemetry/opentelemetry-java/blob/bebf684436b0f54c0c81b19eba1dd9d12ff37532/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogRecordProcessor.java#L61) so it is possible to adjust its behaviour later. There are a couple of issues with reading attrs from each log at this stage though. One is that our [visitor](https://github.com/elastic/apm-agent-android/blob/main/android-sdk/src/main/java/co/elastic/apm/android/sdk/attributes/AttributesVisitor.java#L37) currently only receives an instance of [AttributesBuilder](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/common/AttributesBuilder.java) which doesn't allow reading from it, and the other one is that the log as we get it in the processor is of type [ReadWriteLogRecord](https://github.com/open-telemetry/opentelemetry-java/blob/bebf684436b0f54c0c81b19eba1dd9d12ff37532/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/ReadWriteLogRecord.java) which also doesn't allow to read attrs from itself, although we could call a method it contains to transform it into a [LogRecordData](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogRecordData.java) which is readable, though that process will happen again automatically when sending the log to the exporter, so I'm concerned about duplicating that work.

It would take more work to exclude some logs from having this info than adding it to all, it's not impossible though, we could do a little refactor and provide the existing attrs along with the AttributesVisitor which could enable us to add this logic, or we can even move the visitor to the exporter where we wouldn't have to repeat the work of converting the logs to a readable format. Though in the end what concerns me about this is that, since the UI has this general filter that always relies on the network type info to be available, how can we tell that a similar issue won't arise in the future if we decide to exclude some logs from having it?

bryce-b (approved):

PR #174, opened 2023-08-16
Temporarily store signals in disk 3/3

### LINKED ISSUE #94: Temporarily store signals in disk

LikeTheSalad:
Closes #94 

Final changes to fully add support for disk persistence of signals. I've also updated the docs to include information on how to use this new feature!


docs/configuration.asciidoc
```diff
@@ -104,6 +104,36 @@ class MyApp extends android.app.Application {
...
+                .setExportScheduler(ExportScheduler.getDefault(60 * 60 * 1000)) // One hour by default. The default ExportScheduler will run only when the host app is running, though you can create your own implementation of it in order to provide a better suited scheduler for your app.
+                .build();
+        ElasticApmConfiguration configuration = ElasticApmConfiguration.builder()
+                .setPersistenceConfiguration(persistenceConfiguration)
+                .build();
+        ElasticApmAgent.initialize(this, configuration);
+    }
+}
+----
+
```
bmorelli25, line 136:
```suggestion
                .setEnabled(true) <1>
                .setMaxCacheSize(60 * 1024 * 1024)
                .setExportScheduler(ExportScheduler.getDefault(60 * 60 * 1000))
                .build();
        ElasticApmConfiguration configuration = ElasticApmConfiguration.builder()
                .setPersistenceConfiguration(persistenceConfiguration)
                .build();
        ElasticApmAgent.initialize(this, configuration);
    }
}
----
<1> Defaults to `false`.
<2> Defaults to 60 MB.
<3> Defaults to one hour. The default `ExportScheduler` will run only when the host app is running, though you can create your own implementation of it in order to provide a better-suited scheduler for your app.
```

android-sdk/src/main/java/co/elastic/apm/android/sdk/features/persistence/PersistenceConfiguration.java
```diff
@@ -57,7 +63,19 @@ public Builder setMaxCacheSize(int maxCacheSize) {
...
+         * export it.
+         */
+        public Builder setExportScheduler(ExportScheduler exportScheduler) {
+            this.exportScheduler = exportScheduler;
+            return this;
+        }
+
         public PersistenceConfiguration build() {
+            if (exportScheduler == null) {
+                exportScheduler = ExportScheduler.getDefault(60 * 60 * 1000); // Every hour by default
```
bryce-b, line 77: Is the default for exporting data to the collector 60 minutes? 60 seconds should be more appropriate.

LikeTheSalad: I don't have a strong opinion about this, so if the iOS agent uses 60 seconds already then I'll do the same here to keep consistency!

LikeTheSalad: It's updated now

bryce-b (approved):

PR #179, opened 2023-08-29
Adding session sampling support

### LINKED ISSUE #177: Adding session sampling support

LikeTheSalad:
Closes #177 

These changes add sampling support, configurable through the SDK and also through the central config. I've also updated the docs on how to configure it in both cases.


docs/configuration.asciidoc
```diff
@@ -159,6 +164,35 @@ class MyApp extends android.app.Application {
 }
 ----
 
+[float]
+[[sample-rate-configuration]]
+==== Sample rate configuration
+
+You can set a session sample rate that will be evaluated on every new session to determine whether the full session is exported or ignored.
+Sessions are currently time-based and will kept alive for at least 30 mins, a `session.id` attribute will be sent on every signal until the timer ends, resetting the timer whenever a new signal is created.
```
bmorelli25, line 172:
```suggestion
Sessions are currently time-based and will kept alive for at least 30 mins. A `session.id` attribute will be sent on every signal until the timer ends, resetting the timer whenever a new signal is created.
```

docs/configuration.asciidoc
```diff
@@ -159,6 +164,35 @@ class MyApp extends android.app.Application {
...
+
+When the time's up, a new session ID will be generated and the sample rate will be evaluated to determine whether the new session's signals will get exported or ignored.
+
+You can set the sample rate value at runtime either programmatically, as shown below, or remotely through the <<configuration-dynamic, Central configuration>>.
+The values set through the Central configuration will override the ones set programmatically.
```
bmorelli25, line 177:
```suggestion
Values set through Central configuration will override the ones set programmatically.
```

docs/configuration.asciidoc
```diff
@@ -360,4 +394,21 @@ You can use this setting to dynamically disable Elastic APM at runtime.
...
+|============
+
+[float]
+[[config-session-sample-rate]]
+==== `session_sample_rate` (added[0.9.0])
+
+By default, the agent will sample all signals generated by your application (e.g. spans, metrics, & logs).
+To reduce overhead and storage requirements, you can set the sample rate to a value between 0.0 and 1.0. When reduced below 1.0, data will be sampled per session.
+This is so context in a given session isn't lost.
+You can use this setting to dynamically disable Elastic APM at runtime.
```
bmorelli25, line 406: Should we explain that to do this, you set the sample rate to zero? for example:

```suggestion
You can use this setting to dynamically disable Elastic APM at runtime by setting the sample rate to `0`.
```

docs/configuration.asciidoc
```diff
@@ -360,4 +394,21 @@ You can use this setting to dynamically disable Elastic APM at runtime.
...
+
```
bmorelli25, line 407: We have some good docs in the APM Guide that explain the sample rate in a bit more detail. Should we consider linking to them for more information? https://www.elastic.co/guide/en/apm/guide/current/sampling.html

LikeTheSalad: Thanks for letting me know! I wasn't aware of those and I generally think that the more info we provide the better! However in this case it seems like the docs' sampling rate approach is slightly different from what we do on mobile so adding that link here might cause some confusion. The approach in the docs mentions that a percentage of transactions will be discarded based on the sample rate, but we realized that for mobile it's better to drop all the signals/transactions from a session (or send them all) based on the sample rate so that we wouldn't get incomplete sessions later. Though I think we could at some point create or expand those docs to mention the "session sampling" approach too!

bmorelli25: Ahh, okay makes sense to me!

PR #181, opened 2023-08-30
buildkite: use pre-command to mask secrets

v1v:
I've moved the bits and pieces from the `release.sh` to the `pre-command` hook that's what we did in the past for other APM Agents and that's how we can prepare environment before running anything and mask any secrets in the logs.

Please review it with double eyes


LikeTheSalad (approved):
Thanks! looks good to me.

PR #201, opened 2023-10-16
Fix publishing

LikeTheSalad:
Using armored key

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -68,8 +68,14 @@ export KEYPASS_SECRET
 export KEY_ID_SECRET=D88E42B4
 unset VAULT_TOKEN
 
+echo "--- Listing keys before"
+gpg --list-keys
+
 # Import the key into the keyring
 echo "$KEYPASS_SECRET" | gpg --batch --import "$KEY_FILE"
+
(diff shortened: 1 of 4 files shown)


.buildkite/hooks/pre-command
```diff
@@ -65,18 +65,18 @@ export PLUGIN_PORTAL_SECRET
 vault read -field=key secret/release/signing >$KEY_FILE
 KEYPASS_SECRET=$(vault read -field=passphrase secret/release/signing)
 export KEYPASS_SECRET
-export KEY_ID_SECRET=D88E42B4
```
v1v, line 68: What's the reason?

v1v (approved):

.buildkite/hooks/pre-command
LikeTheSalad: Gradle needed to know the ID of the key to use during the signing process as mentioned [here](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:signatory_credentials). That's why I had to set it up. However, now we've changed to the "[in-memory](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:in-memory-keys)" approach, which doesn't need that value. So since it's no longer needed, I removed it.

PR #240, opened 2023-11-22
Create catalog-info.yaml

esenmarti:
Terrazzo currently uses the data defined in https://github.com/elastic/ci/blob/main/terrazzo/manifests/prod/buildkite/ to define the BK pipeline. Moving forward, teams should store this information in the affected repo so this PR converts the existing data into an RRE and stores it here.


LikeTheSalad (approved):

esenmarti:
Hi @LikeTheSalad ! Could you also merge the PR. I don't think I can. Thanks!

LikeTheSalad:
> Hi @LikeTheSalad ! Could you also merge the PR. I don't think I can. Thanks!

Hi! Sure np 👍

PR #245, opened 2023-12-11
Enhancing internal logging

### LINKED ISSUE #243: Add ability to disable apm agent if its not initialized in code
The logs flooded with 'Configurations have not been initialized.' There should be an option to disable or notify the APM agent that we have intentionally not skipped initialization (e.g., when developing locally or in a staging environment).

LikeTheSalad:
Closes #243 

This PR:

* Disables internal logs for when the Agent is not initialized.
* Provides configuration for customizing the overall internal logs behavior.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -388,12 +388,12 @@ This table outlines the attributes and their corresponding permissions:
 |===
 
 [float]
-==== Configuring this library's logging policy
+==== Logging policy
 
-By default, this library's internal logs are all printed for a debuggable app build.
-In the case of non-debuggable builds, only logs starting from level INFO onwards are printed.
+By default, all logs created by this library are printed for a debuggable app build.
+In the case of non-debuggable builds, only logs at the INFO level and above are printed.
 
-In case you would like to have a more customized policy, or even disable all of the logs from this library altogether, you could do so by providing your own `LoggingPolicy` config like so:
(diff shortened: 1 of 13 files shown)


docs/configuration.asciidoc
```diff
@@ -387,6 +387,32 @@ This table outlines the attributes and their corresponding permissions:
 
 |===
 
+[float]
+==== Configuring this library's logging policy
+
+By default, this library's internal logs are all printed for a debuggable app build.
+In the case of non-debuggable builds, only logs starting from level INFO onwards are printed.
+
+In case you would like to have a more customized policy, or even disable all of the logs from this library altogether, you could do so by providing your own `LoggingPolicy` config like so:
```
bmorelli25, line 396: In general, we try to avoid using gerunds (words ending in `ing`) in section/page headings. "Configure" is also in the page heading, so I don't think we need it here. Can we simplify this to "Logging policy"? Or is that removing too much context?

```suggestion
==== Logging policy

By default, all logs created by this library are printed for a debuggable app build.
In the case of non-debuggable builds, only logs at the INFO level and above are printed.

If you would like to create a custom log policy or even disable all of the logs from this library altogether, you can do so by providing your own `LoggingPolicy` configuration. The below example policy will allow all logs of level WARN and higher to be printed. Levels below WARN will be ignored.
```

LikeTheSalad: Thanks, it makes sense! - For the header, I think we should at least add the word `Internal` at the beginning to emphasize that this is for internal logs only and it has nothing to do with [OTel logs](https://opentelemetry.io/docs/concepts/signals/logs/). Apart from that, I think it's looking great 👍

AlexanderWert (approved):

PR #252, opened 2023-12-22
Removing Gradle plugin module

LikeTheSalad:
Our Gradle plugin was key to providing automatic instrumentation at compile-time, however, now that we're going to change our base project to be [OTel Android](https://github.com/open-telemetry/opentelemetry-android) instead of OTel Java directly, we will no longer need to provide automatic instrumentation from within our Agent, as our users should instead pick the auto-instrumentations they need from the OTel Android [auto-instrumentations module](https://github.com/open-telemetry/opentelemetry-android/tree/main/auto-instrumentation).

A secondary, non crucial benefit, that our Gradle plugin provided was to allow to configure the APM server connectivity params and also the app's info such as `service.name` and `service.version` through Gradle, which, even though it was handy for simple use cases, some more complex apps (such as those with different [variants/flavors](https://developer.android.com/build/build-variants) that might require different config params on each case) might not find it too useful unless it also supported multiple build configs as well. This is a "nice to have" benefit that could be added in the future if the demand for it is high enough, but at the moment, it causes us to keep up with Google's AGP API changes (which tend to happen quite often) so it doesn't seem to be worth it for now.

Based on the above, these changes take care of removing our Gradle plugin module and updating the docs accordingly to better convey our agent's role as an extension of the OTel Android lib.


bryce-b (approved):

docs/faq.asciidoc
```diff
@@ -5,37 +5,26 @@
 [[faq-how-does-it-work]]
 === How does the agent work?
 
-The agent auto-instruments known frameworks and libraries and records interesting events, like HTTP requests.
-To do this, it leverages the capability of the JVM to instrument the bytecode of classes.
-This means that no code changes are required for supported technologies.
+The agent is built on top of https://github.com/open-telemetry/opentelemetry-android[OpenTelemetry Android] and works as an extension that makes it easier to configure and set it all up to work with the Elastic tools with minimum overhead.
```
bmorelli25, line 8:
```suggestion
The agent is built on top of https://github.com/open-telemetry/opentelemetry-android[OpenTelemetry Android]. It works as an extension, making it easier to set up and configure Elastic APM with minimum overhead.
```

docs/faq.asciidoc
```diff
@@ -5,37 +5,26 @@
...
 
-The agent automatically and safely injects small pieces of code before and after interesting events to measure their duration and metadata, like HTTP-related information, including the URL, parameters, and headers.
-
-These events, called Spans, are sent to the APM Server which converts them to a format suitable for Elasticsearch, and sends them to an Elasticsearch cluster.
-You can then use the APM app in Kibana to gain insight into latency issues and error culprits within your application.
-
-The instrumentation happens at compile time, using the https://developer.android.com/reference/tools/gradle-api/7.2/com/android/build/api/variant/Instrumentation[Android Gradle plugin Instrumentation] API.
+All the data created using the agent, either automatically or manually, is sent to the APM Server which converts it to a format suitable for Elasticsearch, and sends them to an Elasticsearch cluster.
```
bmorelli25, line 10:
```suggestion
All data created by the agent, either automatically or manually, is sent to the APM Server, which converts it to a format suitable for Elasticsearch and sends it to an Elasticsearch cluster.
```

docs/faq.asciidoc
```diff
@@ -5,37 +5,26 @@
...
+You can then use the APM app in Kibana to gain insight into latency issues and error culprits within your application and more.
 
 [float]
 [[faq-bytecode-instrumentation]]
 === Is the agent doing bytecode instrumentation?
 
-Yes
+No, although if you'd like to get automatic instrumentations, you can add https://github.com/open-telemetry/opentelemetry-android/tree/main/auto-instrumentation[extensions] to do so that make use of bytecode instrumentation.
```
bmorelli25, line 17:
```suggestion
No, although if you'd like to get automatic instrumentations, you can add https://github.com/open-telemetry/opentelemetry-android/tree/main/auto-instrumentation[extensions] that make use of bytecode instrumentation.
```

docs/faq.asciidoc
```diff
@@ -5,37 +5,26 @@
...
+The metrics aggregation strategy used by the agent is https://github.com/open-telemetry/opentelemetry-java/blob/976edfde504193f84d19936b97e2eb8d8cf060e2/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/AggregationTemporality.java#L15[DELTA].
 
 [float]
 [[faq-unsupported-technologies]]
 === What if the agent doesn't support the technologies I'm using?
 
-The agent is built on top of https://opentelemetry.io/docs/instrumentation/java/getting-started/[OpenTelemetry] You can use its public API to create custom spans and transactions.
\ No newline at end of file
+The agent is built on top of https://opentelemetry.io/docs/instrumentation/java/getting-started/[OpenTelemetry] You can use its public API to create custom spans and transactions.
+You should also take a look at the https://github.com/open-telemetry/opentelemetry-android/tree/main/auto-instrumentation[automatic instrumentation extensions] currently available where the tool you need to instrument might be supported already.
```
bmorelli25, line 30:
```suggestion
The agent is built on top of https://opentelemetry.io/docs/instrumentation/java/getting-started/[OpenTelemetry]. You can use its public API to create custom spans and transactions.
You should also take a look at the https://github.com/open-telemetry/opentelemetry-android/tree/main/auto-instrumentation[automatic instrumentation extensions] currently available where the tool you need to instrument might be supported already.
```

docs/intro.asciidoc
```diff
@@ -1,26 +1,20 @@
...
-This means that for supported technologies, there are no code changes required.
+The Agent is built on top of https://github.com/open-telemetry/opentelemetry-android[OpenTelemetry Android] and helps setting everything up to get your project up and running with Elastic without having to get too deep into the OpenTelemetry API to make it all work.
+It also serves as an extension to provide support for Elastic-specific features, such as <<configuration-dynamic, dynamic configuration>>.
 
-Spans are grouped in transactions -- by default, one for each outgoing HTTP request or UI rendering process.
-It's also possible to create custom transactions, as well as logs and metrics, with the https://opentelemetry.io/docs/instrumentation/java/manual/[OpenTelemetry Java API], which is automatically provided to the Agent's host app.
+It makes sure to configure OpenTelemetry in a way that any custom Spans, as well as Logs and Metrics, created using the https://opentelemetry.io/docs/instrumentation/java/manual/[OpenTelemetry Java API], contain useful attributes that enrich the overall telemetry data.
```
bmorelli25, line 14: We shouldn't document this in two places. I would remove the "How does the agent work" section from the FAQ and only have this documented here.

docs/manual-instrumentation.asciidoc
```diff
@@ -1,8 +1,7 @@
 [[manual-instrumentation]]
 == Manual Instrumentation
 
-The Elastic APM Android Agent automatically instruments <<supported-technologies>>, creating spans for interesting events for each case, and some of those automated spans can be <<configuration,configured>> to better suit different app's needs.
-However, if you need to create your own, custom spans, metrics and logs, you can do so by accessing the https://opentelemetry.io/docs/instrumentation/java/manual/[OpenTelemetry Java SDK APIs] that this agent is built on top.
+If you need to create your own, custom spans, metrics and logs, you can do so by accessing the https://opentelemetry.io/docs/instrumentation/java/manual/[OpenTelemetry Java SDK APIs] that this agent is built on top.
```
bmorelli25, line 4:
```suggestion
If you need to create your own custom spans, metrics, and logs, you can do so by accessing the https://opentelemetry.io/docs/instrumentation/java/manual/[OpenTelemetry Java SDK APIs] that this agent is built on top of.
```

docs/setup.asciidoc
```diff
@@ -83,12 +68,24 @@ class MyApp extends android.app.Application {
...
+        // Optionally if you also define a custom configuration:
+        // ElasticApmAgent.initialize(this, ElasticApmConfiguration.builder().build(), myConnectivityConfig);
     }
 }
 ----
 
-<1> Initialize the Elastic APM agent once.
+<1> Connects to an APM server that requires no authentication.
+<2> Connects to an APM server that requires an https://www.elastic.co/guide/en/apm/guide/current/api-key.html[API KEY] for authentication.
+<3> Connects to an APM server that requires a https://www.elastic.co/guide/en/apm/guide/current/secret-token.html[SECRET TOKEN] for authentication.
```
bmorelli25, line 85:
```suggestion
<2> Connects to an APM server that requires an {apm-guide-ref}/api-key.html[API key] for authentication.
<3> Connects to an APM server that requires a {apm-guide-ref}/secret-token.html[secret token] for authentication.
```

docs/setup.asciidoc
```diff
@@ -97,24 +94,17 @@ class MyApp extends android.app.Application {
...
-NOTE: Without the Gradle plugin, the Elastic SDK won't be able to provide automatic instrumentations for its <<supported-technologies>>.
-
 [float]
 [[whats-next]]
 === What's next?
 
-After initializing the agent (by using the gradle plugin), your application will automatically create traces for all OkHttp network requests (including those created by tools that make use of OkHttp, like Retrofit) and all https://developer.android.com/reference/android/app/Activity[Activity] and https://developer.android.com/reference/androidx/fragment/app/Fragment[Fragment] starting methods.
+After initializing the agent, your application will:
+
+- Create traces for all https://developer.android.com/reference/android/app/Activity[Activity] and https://developer.android.com/reference/androidx/fragment/app/Fragment[Fragment] classes' starting methods.
```
bmorelli25, line 103:
```suggestion
- Create traces for all https://developer.android.com/reference/android/app/Activity[Activity] and https://developer.android.com/reference/androidx/fragment/app/Fragment[Fragment] classes starting methods.
```

bmorelli25 (commented):
Nice improvements. I have a few suggestions for you.

docs/intro.asciidoc
LikeTheSalad: Makes sense! I've just removed it from the FAQ doc 👍

PR #282, opened 2024-03-14
Cicd version validation

LikeTheSalad:


build-tools/src/main/java/co/elastic/apm/compile/tools/publishing/ApmPublisherRootPlugin.java
```diff
@@ -35,12 +37,29 @@ public void apply(Project project) {
...
         if (versionOverride != null) {
+            validateVersionOverrideFormatting(versionOverride);
+            validateVersionOverrideValues(project.getVersion().toString(), versionOverride);
             System.out.println("Overriding version with: '" + versionOverride + "'");
             project.setVersion(versionOverride);
             project.subprojects(subproject -> subproject.setVersion(versionOverride));
         }
     }
 
+    private static void validateVersionOverrideValues(String currentVersion, String versionOverride) {
```
v1v, line 48: IIUC, this function validates that only `patches` are allowed. If so, could the function's name reflect that? Or do you envision to support some other cases in the future?

v1v (approved):
My comment was just a nit, other than that LGTM

build-tools/src/main/java/co/elastic/apm/compile/tools/publishing/ApmPublisherRootPlugin.java
LikeTheSalad: Got it, I see what you mean. I had an idea of allowing also lower numbers in case we wanted to release a patch version for a very old one, but now that I think about it better, I guess in that case we'd create a branch off of the version tag which will have the old version set in `gradle.properties` file already, essentially just making this parameter to still validate only patch number changes are allowed anyway. I'll update it to make it clearer, cheers!

LikeTheSalad: It should be better now 👍

v1v (approved):

build-tools/src/main/java/co/elastic/apm/compile/tools/publishing/ApmPublisherRootPlugin.java
```diff
@@ -35,12 +37,29 @@ public void apply(Project project) {
...
             project.setVersion(versionOverride);
             project.subprojects(subproject -> subproject.setVersion(versionOverride));
         }
     }
 
+    private static void validateVersionPatchChangeOnly(String currentVersion, String versionOverride) {
+        VersionNumber comparableVersion = VersionNumber.parse(currentVersion);
+        VersionNumber comparableVersionOverride = VersionNumber.parse(versionOverride);
+        if (comparableVersionOverride.getMajor() > comparableVersion.getMajor() || comparableVersionOverride.getMinor() > comparableVersion.getMinor()) {
+            throw new IllegalArgumentException(String.format("The version override, '%s', cannot provide greater major or minor numbers than the existing version from the gradle.properties file: '%s'.", versionOverride, currentVersion));
```
estolfo, line 52: Can you explain what you're checking here? Is it making sure that the new version's minor number is not greater than the existing version's minor number? In that case, how do you release a new minor?

LikeTheSalad: > Can you explain what you're checking here? Is it making sure that the new version's minor number is not greater than the existing version's minor number?

Yes, checking that neither the major nor the minor version numbers are greater than the project's next version, which is located [here](https://github.com/elastic/apm-agent-android/blob/main/gradle.properties#L5). This code only runs when releasing a patch version, in which case we would manually pass the patch version in the release Github action parameters which will override the existing project's version. However, since it's a free-format text right now, I'm adding this validation to make sure we won't accidentally release a new minor or major version number.

> how do you release a new minor?

The normal release process is unchanged in this PR, currently, the next minor version is the one defined as the project's version [here](https://github.com/elastic/apm-agent-android/blob/main/gradle.properties#L5). So when we run a normal release, without overriding the project's next version, the project's version will get used in the generated artifact. After the release succeeds, we automatically create a PR, [like this one](https://github.com/elastic/apm-agent-android/pull/277), where we bump the minor version for the next release.

estolfo (approved):

PR #310, opened 2024-05-08
Adding faq on SSL/TLS config

LikeTheSalad:


trentm (dismissed):

docs/faq.asciidoc
```diff
@@ -34,6 +34,47 @@ It is created by a Java Champion, awarded with the Dukes Choice award and curren
...
+=== How can I configure SSL/TLS?
+
+Please note that the Elastic Agent does not handle SSL/TLS configs internally, therefore, the recommended way to manage these types of configurations is by doing so as part of your app's network security configurations, as explained in Android's official https://developer.android.com/privacy-and-security/security-ssl[security guidelines].
+Below we show a set of common use-cases and quick tips on what could be done on each one, however, each case might be different, so please refer to Android's https://developer.android.com/privacy-and-security/security-config[official docs] on this topic in case you need more details.
+
+[float]
+[[faq-ssl-elastic-cloud]]
+==== Connecting to Elastic Cloud
+
+If your ELK stack is hosted in Elastic Cloud, you shouldn't need to add any SSL/TLS config changes in your app, it should work out of the box.
```
bmorelli25, line 48: We don't use ELK in our in docs anymore.

```suggestion
If your Elastic Stack is hosted in Elastic Cloud, you shouldn't need to add any SSL/TLS config changes in your app, it should work out of the box.
```

docs/faq.asciidoc
```diff
@@ -34,6 +34,47 @@ It is created by a Java Champion, awarded with the Dukes Choice award and curren
...
+
+[float]
+[[faq-ssl-on-prem]]
+==== Connecting to an on-prem server
+
+If your ELK stack is hosted on-prem, then it depends on the type of CA your host uses to sign its certificates, if it's a commonly trusted CA, then you shouldn't have to worry about changing your app's SSL/TLS configuration as it all should work well out of the box, however, if your CAs are unknown/private or your server uses a self-signed certificate, then you would need to configure your app to trust custom CAs by following https://developer.android.com/privacy-and-security/security-config[Android's guide on it].
```
bmorelli25, line 54:
```suggestion
If your Elastic Stack is hosted on-prem, then it depends on the type of CA your host uses to sign its certificates, if it's a commonly trusted CA, then you shouldn't have to worry about changing your app's SSL/TLS configuration as it all should work well out of the box, however, if your CAs are unknown/private or your server uses a self-signed certificate, then you would need to configure your app to trust custom CAs by following https://developer.android.com/privacy-and-security/security-config[Android's guide on it].
```

docs/faq.asciidoc
```diff
@@ -34,6 +34,47 @@ It is created by a Java Champion, awarded with the Dukes Choice award and curren
...
+
+[float]
+[[faq-ssl-debug]]
+==== Debugging purposes
+
+If you're running a local server and need to connect to it without using https in order to run a quick test, then you could temporarily https://developer.android.com/guide/topics/manifest/application-element#usesCleartextTraffic[enable clearttext traffic] within your `AndroidManifest.xml` file, inside the `<application>` tag.
```
bmorelli25, line 60:
```suggestion
If you're running a local server and need to connect to it without using https in order to run a quick test, then you could temporarily https://developer.android.com/guide/topics/manifest/application-element#usesCleartextTraffic[enable cleartext traffic] within your `AndroidManifest.xml` file, inside the `<application>` tag.
```

bmorelli25 (approved):

PR #316, opened 2024-05-17
Fetching remote branches before switching to version branch

LikeTheSalad:
Should fix this issue: https://github.com/elastic/apm-agent-android/actions/runs/9129073287/job/25102761808

### DIFF
--- a/.github/workflows/updateVersionBranch.yml
+++ b/.github/workflows/updateVersionBranch.yml
@@ -28,6 +28,8 @@ jobs:
           email: ${{ env.GIT_EMAIL }}
           token: ${{ env.GITHUB_TOKEN }}
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - uses: actions-ecosystem/action-regex-match@v2
         id: major-version
         with:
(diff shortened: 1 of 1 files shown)


trentm (approved):
LGTM.

Probably overkill. The `--all` means to fetch from all remotes... and I assume the only remote from `actions/checkout` is `origin`, so `git fetch` would accomplish the same. Or could use: https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches

LikeTheSalad:
> LGTM.
> 
> Probably overkill. The `--all` means to fetch from all remotes... and I assume the only remote from `actions/checkout` is `origin`, so `git fetch` would accomplish the same. Or could use: https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches

Yeah, it's a bit overkill. I wasn't aware of the existing action's param for this use case, though it seems to be the cleanest approach, so I just made the changes to use it instead. Cheers!

v1v (approved):
That's what we use in other places. LGTM

PR #357, opened 2024-09-17
ci: use GitHub app for ephemeral tokens

v1v:
#### What

Use the GitHub app to generate the required ephemeral tokens with the least permissive principle.

#### Why

* Finer-grained tokens with Service Machine accounts are required to rotate the secrets manually.
* GitHub app to generate temporary tokens is the advanced approach to avoid the above
* Document what the GH workflow requires to run in terms of access
* GitHub Token with Permissions does not trigger GitHub builds 

#### Implementation details

Use `tibdex/github-app-token` with the required permissions and the repository scope
Remove the `permissions` configuration in the GH workflow.
Configure git checkout with the ephemeral token
Configure the GH_TOKEN with the ephemeral token

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -110,16 +110,30 @@ jobs:
     name: Post Release
     needs: release
     runs-on: ubuntu-latest
-    permissions:
-      contents: write
-    env:
-      GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
     steps:
+      - name: Get token
+        id: get_token
+        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
+        with:
+          app_id: ${{ secrets.OBS_AUTOMATION_APP_ID }}
+          private_key: ${{ secrets.OBS_AUTOMATION_APP_PEM }}
+          permissions: >-
+            {
+              "contents": "write",
+              "pull_requests": "write"
+            }
+          repositories: >-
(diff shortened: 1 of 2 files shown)


reakaleek (dismissed):

.github/workflows/release.yml
```diff
@@ -110,16 +110,30 @@ jobs:
...
+        with:
+          app_id: ${{ secrets.OBS_AUTOMATION_APP_ID }}
+          private_key: ${{ secrets.OBS_AUTOMATION_APP_PEM }}
+          permissions: >-
+            {
+              "contents": "write",
+              "pull_requests": "write"
+            }
+          repositories: >-
+            ["${{github.repository}}"]
```
v1v, line 126: See https://github.com/tibdex/github-app-token/issues/104
```suggestion
            ["apm-agent-android"]
```

.github/workflows/updateVersionBranch.yml
```diff
@@ -9,21 +9,36 @@ on:
...
+        with:
+          app_id: ${{ secrets.OBS_AUTOMATION_APP_ID }}
+          private_key: ${{ secrets.OBS_AUTOMATION_APP_PEM }}
+          permissions: >-
+            {
+              "contents": "write",
+              "pull_requests": "write"
+            }
+          repositories: >-
+            ["${{github.repository}}"]
```
v1v, line 32: See https://github.com/tibdex/github-app-token/issues/104
```suggestion
            ["apm-agent-android"]
```

PR #397, opened 2025-02-24
Remoteconfig url param

LikeTheSalad:


bryce-b (approved):
Looks good! I'll use these same configuration options on this iOS agent for consistence. Although, I'll provide a deprecation period for the `url` param, since iOS is already GA.

PR #443, opened 2025-04-01
Updating legacy doc

LikeTheSalad:
Temporary solution

Alternative to avoid old CI checks issues: https://github.com/elastic/apm-agent-android/pull/444

PR #444, opened 2025-04-01
Main legacy docs changes

LikeTheSalad:
As a temporary solution

This is an alternative to https://github.com/elastic/apm-agent-android/pull/443


colleenmcginnis (changes requested):
Use https://github.com/elastic/apm-agent-android/pull/443 instead.

PR #508, opened 2025-06-04
Prepare release notes for migration

theletterf:
This prepares the Android release notes for https://github.com/elastic/opentelemetry-dev/issues/923

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+*.md @elastic/ingest-docs 
+/docs/ @elastic/ingest-docs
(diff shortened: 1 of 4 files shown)


theletterf:
@LikeTheSalad Ready for approval! Could you add the 1.1.0 changes to the PR?

LikeTheSalad:
> @LikeTheSalad Ready for approval! Could you add the 1.1.0 changes to the PR?

Do you mean [these changes](https://github.com/elastic/apm-agent-android/pull/524), @theletterf?

theletterf:
Correct! Also, it seems we need to fix the CODEOwners file.

LikeTheSalad:
> Correct! Also, it seems we need to fix the CODEOwners file.

Got it. Regarding the release notes, it seems like they can get merged without issues, as the new changes are already in `main`, so there's no need to add the same changes in this PR. Apart from that, is there any other changes you think are missing here?

Issue #500, opened 2025-05-27
Provide Builder config to disable disk cache


almozavr:
Since v. `1.0.0` there is no way to disable disk cache, as it's always enabled in the `ManagedFeatures.Builder`. Unfortunately, we can't update to `1.0.0` because of this.

almozavr:
as a tmp solution

```kotlin
private fun ElasticApmAgent.Builder.setDiskBufferingConfiguration(configuration: DiskBufferingConfiguration): ElasticApmAgent.Builder =
    this.apply {
        val diskConfiguration = this::class.declaredMemberFunctions.find { it.name == "setDiskBufferingConfiguration" }
        diskConfiguration?.isAccessible = true
        diskConfiguration?.call(this, configuration)
    }
```

PR #509, opened 2025-06-05
Disk buffering config setter

LikeTheSalad:
Addresses #500


docs/reference/configuration.md
```diff
@@ -81,6 +81,25 @@ class MyApp : android.app.Application {
...
+
+You can configure how [disk buffering](index.md#disk-buffering) works. For example:
+
+```kotlin
+class MyApp : android.app.Application {
+
+    override fun onCreate() {
+        super.onCreate()
+        val agent = ElasticApmAgent.builder(this)
+            .setDiskBufferingConfiguration(DiskBufferingConfiguration.enabled()) // <1>
```
theletterf, line 94: The example here describes how to enable/disable disk buffering. Do we plan on adding more settings here in the future?

PR #653, opened 2025-10-08
Docs update

LikeTheSalad:


README.md
```diff
@@ -15,4 +15,4 @@ In order to use a local build for your local projects, you'll need to publish it
 maven local repo. To do so, open up a terminal in this project's root dir and
 run: `./gradlew publishToMavenLocal`. After that, you can apply the agent into an Android
 application project (that uses [mavenLocal()](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.artifacts.dsl/-repository-handler/maven-local.html) as one of its dependency repositories)
-by following the [Getting started](https://www.elastic.co/guide/en/apm/agent/android/current/setup.html)" guide and using the version defined [here](gradle.properties).
+by following the [Getting started](https://www.elastic.co/guide/en/apm/agent/android/current/setup.html) guide and using the version defined [here](gradle.properties).
```
theletterf, line 18: @LikeTheSalad Could we update this to the new version?

docs/reference/edot-android/automatic-instrumentation.md
```diff
@@ -129,3 +129,28 @@ dependencies {  // <2>
 1. Make sure the adapter is added.
 2. You can find the dependencies needed in the [instrumentation's README file](https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation/httpurlconnection#project-dependencies). The same will be the case for any other instrumentation.
 3. The instrumentations that require a `byteBuddy` dependency, do bytecode weaving, as explained in [compilation behavior](#compilation-behavior). An extra plugin named `net.bytebuddy.byte-buddy-gradle-plugin` is required to make this work, as shown [here](https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation/httpurlconnection#byte-buddy-compilation-plugin). However, EDOT Android installs this extra plugin on your behalf, so there's no need for you to do so manually.
+
+### Compilation performance
+
+As explained in [compilation behavior](#compilation-behavior), the instrumentations that perform bytecode weaving can increase the
```
theletterf, line 135:
```suggestion
As explained in [compilation behavior](#compilation-behavior), the instrumentations that perform bytecode weaving might increase the
```

docs/reference/edot-android/automatic-instrumentation.md
```diff
@@ -129,3 +129,28 @@ dependencies {  // <2>
...
+compilation times, which might not be needed all the time. Because of this, EDOT Android provides a [configuration param](#automatic-instrumentation-configuration)
```
theletterf, line 136: > which might not be needed all the time.

Bytecode weaving is not needed all the time? If that's the case, I suggest this rewrite:

```suggestion
compilation times. Because bytecode weaving might not be necessary, EDOT Android provides a [configuration param](#automatic-instrumentation-configuration)
```

docs/reference/edot-android/automatic-instrumentation.md
```diff
@@ -129,3 +129,28 @@ dependencies {  // <2>
...
+that allows to select specific build types for which the bytecode instrumentation will be disabled. However, said configuration
```
theletterf, line 137:
```suggestion
that allows to select specific build types for which the bytecode instrumentation will be turned off. This configuration
```

docs/reference/edot-android/automatic-instrumentation.md
```diff
@@ -129,3 +129,28 @@ dependencies {  // <2>
...
+only works for EDOT Android's [supported instrumentations](#supported-instrumentations), so it won't have any effect on
+instrumentations added via the OTel Android instrumentations adapter.
```
theletterf, line 139:
```suggestion
instrumentations added through the OTel Android instrumentations adapter.
```

docs/reference/edot-android/automatic-instrumentation.md
```diff
@@ -129,3 +129,28 @@ dependencies {  // <2>
...
+
+In order to select the build variant to run the bytecode instrumentation for an OTel Android instrumentation, you must
+add its `byteBuddy` dependency using the variant-specific byteBuddy name, such as `releaseByteBuddy` or `debugByteBuddy`,
+depending on which variant you'd like to enable the bytecode instrumentation for.
+
+Following the example above, if we'd like to install the `HttpURLConnection` instrumentation only for our app's `release` build type,
+we can change the way we add its byteBuddy dependency as shown below:
```
theletterf, line 146:
```suggestion
Following the previous example, if you want to install the `HttpURLConnection` instrumentation only for your app's `release` build type,
you can change the way we add its byteBuddy dependency as shown here:
```

docs/reference/edot-android/getting-started.md
```diff
@@ -27,7 +27,8 @@ Set up the Elastic Distribution of OpenTelemetry Android (EDOT Android) in your
 | Android API level                                 | 26 (or 21 with [desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring)) |
 
 :::{important}
-If your application's [minSdk](https://developer.android.com/studio/publish/versioning#minsdk) value is lower than 26, you must add [Java 8 desugaring support](https://developer.android.com/studio/write/java8-support#library-desugaring). Refer to [Troubleshooting](docs-content://troubleshoot/ingest/opentelemetry/edot-sdks/android/index.md#why-desugaring) for more information.
+* If your application's [minSdk](https://developer.android.com/studio/publish/versioning#minsdk) value is lower than 26, you must add [Java 8 desugaring support](https://developer.android.com/studio/write/java8-support#library-desugaring). Refer to [Troubleshooting](docs-content://troubleshoot/ingest/opentelemetry/edot-sdks/android/index.md#why-desugaring) for more information.
+* EDOT Android is meant to be used in Android projects only. Hybrid frameworks, such as React Native and Flutter, are not supported.
```
theletterf, line 31: Very important note!

docs/reference/edot-android/index.md
```diff
@@ -80,7 +80,7 @@ For [distributed tracing](#distributed-tracing) to work properly, your applicati
 
 A session groups telemetry data within a time frame when your application is active by adding a `session.id` attribute to all the spans and logs that come out of your application. This helps limit the volume of data you need to look at when investigating the steps that led to an error within your app. It can also give you insight into the common actions performed on each use.
 
-A session is created when no previous session exists or when the previous one has expired. Sessions expire after 30 minutes of inactivity. You can extend this for up to 4 hours.
+A session is created when no previous session exists or when the previous one has expired. Sessions expire after 30 minutes of inactivity. If there's constant activity on an app, the session will live up to 4 hours maximum.
```
theletterf, line 83:
```suggestion
A session is created when no previous session exists or when the previous one has expired. Sessions expire after 30 minutes of inactivity. If there's constant activity on an app, the session will live up to a maximum of 4 hours.
```

theletterf (changes requested):
Nice! Left some suggestions.

docs/reference/edot-android/automatic-instrumentation.md
LikeTheSalad: > Bytecode weaving is not needed all the time?

Well, maybe the wording I used is not the right one now that I think about it. It's always necessary to make the instrumentation work properly, but it's just that adding observability to an app might not always be necessary (like during development for example), and rather is more of a tool for apps in production. And since developers tend to build an app multiple times while adding code changes to check how their new code behaves, it might be annoying to slow down the compilation in these cases because of a feature that they might not want to use during development, such as observability. Let me see how I can rephrase it a little bit.

README.md
LikeTheSalad: This is a great point, it's updated now 👍

theletterf (approved):
Yay! Let's go :)

docs/reference/edot-android/automatic-instrumentation.md
LikeTheSalad: I've just removed the confusing part. I noticed that the message I wanted to convey is still there without that part, and it's more straightforward to read as well.

PR #662, opened 2025-10-15
Update release branch

LikeTheSalad:
Needed to fix maven central release

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -34,7 +34,7 @@ echo "--- Prepare keys context :key:"
 set +x
 # Nexus credentials (they cannot use the _SECRET pattern since they are in-memory based)
 # See https://docs.gradle.org/current/userguide/signing_plugin.html#sec:in-memory-keys
-NEXUS_SECRET=kv/ci-shared/release-eng/team-release-secrets/apm/maven_central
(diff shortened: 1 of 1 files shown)


alina-bacalete (approved):
LGTM

.buildkite/hooks/pre-command
```diff
@@ -34,7 +34,7 @@ echo "--- Prepare keys context :key:"
 set +x
 # Nexus credentials (they cannot use the _SECRET pattern since they are in-memory based)
 # See https://docs.gradle.org/current/userguide/signing_plugin.html#sec:in-memory-keys
-NEXUS_SECRET=kv/ci-shared/release-eng/team-release-secrets/apm/maven_central
+NEXUS_SECRET=kv/ci-shared/release-eng/team-release-secrets/otel/maven_central
```
v1v: I'm not sure if this is the way, I was told that the namespace `otel` should be only for the otel agents here.

LikeTheSalad: I'm guessing that with otel agents in this context, we're talking about EDOT agents/sdks? Because if that's the case, this is an otel agent too. It's just that the repo predates the creation of the EDOT term, which is why the repo name still reads `apm`.

v1v: Gotcha, thanks for the clarification!

PR #742, opened 2026-02-27
Update CI integration tests

LikeTheSalad:
This PR enhances the integration tests to allow for checking an app with and without desugaring (previously only checked without desugaring). It also makes the integration tests rely on start-local for setting up the Elasticsearch + EDOT Collector (instead of using start-local just for Elasticsearh, and then using a custom script to spin up the collector, as it is right now)


fr4nc1sc0-r4m0n (approved):
LGTM

PR #877, opened 2026-08-13
Android r8 mapping upload

LikeTheSalad:


LikeTheSalad:
Hi @lucabelluccini - I was told that you might be interested in checking the doc changes from this PR. It's focused on finishing the Android crash deobfuscation work.

agent-plugin/src/main/java/co/elastic/otel/android/plugin/internal/mapping/upload/UploadMappingToElasticsearch.kt
```diff
@@ -0,0 +1,234 @@
...
+            .url("$base/$indexName")
+            .put(body.toRequestBody(JSON_MEDIA_TYPE))
+            .header("Authorization", "ApiKey ${apiKey.get()}")
+            .build()
+
+        client.newCall(request).execute().use { response ->
+            if (response.isSuccessful) {
+                logger.lifecycle("Created mapping index: $indexName")
+            } else {
+                val responseBody = response.body.string()
```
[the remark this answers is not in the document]
LikeTheSalad: Thanks for flagging this. In the OkHttp 5.4.0 API used here, `Response.body` is non-null (`val body: ResponseBody`) and the response builder defaults it to `ResponseBody.EMPTY`, so a safe call is not needed. We will keep the current code.

.github/scripts/integration-test/integration_test.sh
```diff
@@ -39,7 +39,7 @@ launch_app() {
   if [ "${WITH_DESUGARING:-false}" = "true" ]; then
     gradle_args+=("-PwithDesugaring=true")
   fi
-  "$app_dir/gradlew" -p "$app_dir" :app:assembleRelease "${gradle_args[@]}"
+  "$app_dir/gradlew" -p "$app_dir" :app:assembleRelease ${gradle_args[@]+"${gradle_args[@]}"}
```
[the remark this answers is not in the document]
LikeTheSalad: The guarded expansion is intentional for compatibility with macOS Bash 3.2. This script enables `set -u`, and a normal `"${gradle_args[@]}"` expansion fails with an unbound-variable error when the array is empty. `${gradle_args[@]+"${gradle_args[@]}"}` safely passes no arguments in that case, so we will keep it.

agent-plugin/src/main/java/co/elastic/otel/android/plugin/extensions/ElasticMappingExtension.kt
```diff
@@ -0,0 +1,31 @@
...
+ * under the License.
+ */
+package co.elastic.otel.android.plugin.extensions
+
+import javax.inject.Inject
+import org.gradle.api.Action
+import org.gradle.api.model.ObjectFactory
+
+abstract class ElasticMappingExtension @Inject constructor(objects: ObjectFactory) {
+    val elasticsearch: Elasticsearch = objects.newInstance(Elasticsearch::class.java)
```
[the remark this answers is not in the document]
LikeTheSalad: Gradle DSL extension types created through `ObjectFactory` are not generally required to implement `Serializable`. AGP recommends that the top-level `VariantExtension` subtype be serializable, which `ElasticVariantExtension` already is; it does not require every nested DSL type to declare it. Only the nested `Property` values are wired into tasks, and strict configuration-cache storage and reuse succeeds with this implementation, so there is no serialization issue to address here.

agent-plugin/src/main/java/co/elastic/otel/android/plugin/extensions/Elasticsearch.kt
```diff
@@ -0,0 +1,26 @@
...
+ * under the License.
+ */
+package co.elastic.otel.android.plugin.extensions
+
+import org.gradle.api.provider.Property
+
+interface Elasticsearch {
+    val endpoint: Property<String>
+    val apiKey: Property<String>
+}
```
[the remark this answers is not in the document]
LikeTheSalad: Gradle DSL extension types created through `ObjectFactory` are not generally required to implement `Serializable`. AGP recommends that the top-level `VariantExtension` subtype be serializable, which `ElasticVariantExtension` already is; it does not require every nested DSL type to declare it. Only the nested `Property` values are wired into tasks, and strict configuration-cache storage and reuse succeeds with this implementation, so there is no serialization issue to address here.

agent-plugin/src/main/java/co/elastic/otel/android/plugin/internal/mapping/upload/GenerateElasticsearchBulkRequestBody.kt
```diff
@@ -0,0 +1,71 @@
...
+    @get:OutputFile
+    abstract val requestBodyFile: RegularFileProperty
+
+    @TaskAction
+    fun generate() {
+        val mapping = mappingFile.orNull?.asFile
+        if (mapping == null || !mapping.exists()) {
+            logger.lifecycle("No mapping file found. Skipping bulk request body generation.")
+            return
+        }
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I just addressed it here: https://github.com/elastic/apm-agent-android/pull/877/commits/4c06e4a3fda0a4554610f6acd0fbb98b4bf32ce3

jackshirazi (dismissed):

docs/reference/edot-android/crash-reporting.md
```diff
@@ -0,0 +1,180 @@
...
+android {
+    buildTypes {
+        release {
+            isMinifyEnabled = true
+        }
+    }
+}
+```
+> [!NOTE]
+> The endpoint must be the {{es}} HTTP endpoint, not the OTLP endpoint used to export telemetry.
```
alexandra5000:
```suggestion
:::{note}
The endpoint must be the {{es}} HTTP endpoint, not the OTLP endpoint used to export telemetry.
:::
```

docs/reference/edot-android/crash-reporting.md
```diff
@@ -0,0 +1,180 @@
...
+    ]
+  }
+}
+```
+
+5. Set an expiration that matches your release process, create the key, and store its encoded value as the `ELASTICSEARCH_API_KEY` CI secret.
+
+This API key is a build-time credential. Don't package it in the application or commit it to source control.
+
+The user creating the key must have permission to create API keys and the index privileges being granted. Refer to [Elastic API keys](https://www.elastic.co/docs/deploy-manage/api-keys) for deployment-specific requirements and instructions.
```
alexandra5000:
```suggestion
The user creating the key must have API key creation permissions and must already hold the index privileges being granted. Refer to [Elastic API keys](docs-content://deploy-manage/api-keys) for deployment-specific requirements and instructions.
```

Rephrased for clarity/brevity

docs/reference/edot-android/crash-reporting.md
```diff
@@ -0,0 +1,180 @@
...
+The instrumentation automatically captures crashes when an unhandled exception occurs. Because the application process is terminating, EDOT Android stores the crash event on disk and exports it after EDOT Android initializes the next time the application is launched.
+
+Crash events are available in the "Crashes" section of the {{kib}} Android dashboard. Refer to [Visualize your telemetry](getting-started.md#visualize-telemetry) for instructions to install and open the dashboard.
+
+## Deobfuscate R8 stacktraces
+
+```{applies_to}
+product:
+  edot_android: ga 1.8.0
+```
```
alexandra5000, line 43:
```{applies_to}
stack: ga 9.3.9
product:
  edot_android: ga 1.8.0
```

If I understand it correctly, deobfuscation also requires specific Kibana versions (8.19.20, 9.3.9, 9.4.5, or 9.5.0+), because it depends on the Android OpenTelemetry Assets integration v1.0.0+. Without a stack: entry in these applies_to blocks, users won't see any Stack version badge and may assume the feature works on any Stack version. I'm putting 9.3.9 in there as the oldest 9.x supported version. 

Alternatively, if you think the tags would be confusing here or only some versions are supported and it wouldn't make sense to tag them all, we could add 1 sentence in prose with a link to: https://www.elastic.co/docs/reference/integrations/otel_android_dashboards

docs/reference/edot-android/configuration.md
```diff
@@ -332,6 +332,17 @@ android {
 
 When multiple values match a variant, EDOT Android uses the most specific value in this order: build type, product flavor, project-level `elasticOtel`, then the generated default.
 
+### R8 mapping upload
+
+```{applies_to}
+product:
+  edot_android: ga 1.8.0
+```
```
alexandra5000, line 340:
```{applies_to}
stack: ga 9.3.9
product:
  edot_android: ga 1.8.0
```

If I understand it correctly, deobfuscation also requires specific Kibana versions (8.19.20, 9.3.9, 9.4.5, or 9.5.0+), because it depends on the Android OpenTelemetry Assets integration v1.0.0+. Without a stack: entry in these applies_to blocks, users won't see any Stack version badge and may assume the feature works on any Stack version. I'm putting 9.3.9 in there as the oldest 9.x supported version. 

Alternatively, if you think the tags would be confusing here or only some versions are supported and it wouldn't make sense to tag them all, we could add 1 sentence in prose with a link to: https://www.elastic.co/docs/reference/integrations/otel_android_dashboards

docs/reference/edot-android/crash-reporting.md
```diff
@@ -0,0 +1,180 @@
...
+```bash
+./gradlew paidReleaseUploadMappingToElasticsearch
+```
+
+The upload task:
+
+1. Ensures R8 has produced `mapping.txt` for the selected variant.
+2. Converts the mapping into one {{es}} document per obfuscated class.
+3. Creates `.android-r8-mappings-<build_id>` if it doesn't already exist.
+4. Uploads the documents using the {{es}} Bulk API.
```
alexandra5000, line 164: Are these things done in that order by the upload task or does the order matter in some other way? If not, consider a simple bullet list rather than the numbered list.

docs/reference/edot-android/crash-reporting.md
```diff
@@ -0,0 +1,180 @@
...
+
+```kotlin
+plugins {
+    id("com.android.application")
+    id("co.elastic.otel.android.agent") version "[latest_version]"
+    id("co.elastic.otel.android.mapping") version "[latest_version]" // <1>
+}
+```
+
+1. Find the latest mapping plugin version in the [Gradle Plugin Portal](https://plugins.gradle.org/plugin/co.elastic.otel.android.mapping).
```
alexandra5000:
```suggestion
Find the latest mapping plugin version in the [Gradle Plugin Portal](https://plugins.gradle.org/plugin/co.elastic.otel.android.mapping).
```

Removing as there is no point number 2 (so no point adding this as one, just keep it a regular sentence)

docs/reference/edot-android/crash-reporting.md
```diff
@@ -0,0 +1,180 @@
...
+Android's R8 optimizer can rename classes and methods in release builds. When an optimized application crashes, its stacktrace contains these obfuscated names. EDOT Android can upload the R8 `mapping.txt` file for each application build to {{es}} so that the stacktrace can be restored to its original class, method, file, and line information.
+
+EDOT Android identifies the correct mapping using the `app.build_id` resource attribute included with application telemetry. Mapping documents for a build are stored in a {{es}} index named `.android-r8-mappings-<build_id>`.
+
+### Prerequisites
+
+* Enable R8 for the variant whose mapping you want to upload.
+* Have your {{es}} endpoint URL at hand to use it in the Gradle task later.
+* Create a dedicated {{es}} API key for mapping uploads.
+* Install version 1.0.0 or later of the **Android OpenTelemetry Assets** integration in {{kib}}. Refer to [Visualize telemetry](getting-started.md#visualize-telemetry) for instructions on how to do so.
```
alexandra5000: Isn't the Kibana version a prerequisite as well, as per https://www.elastic.co/docs/reference/integrations/otel_android_dashboards ?

docs/reference/edot-android/crash-reporting.md
```diff
@@ -0,0 +1,180 @@
...
+
+```kotlin
+plugins {
+    id("com.android.application")
+    id("co.elastic.otel.android.agent") version "[latest_version]"
+    id("co.elastic.otel.android.instrumentation.crash") version "[latest_version]" // <1>
+}
+```
+
+1. Find the latest crash reporting plugin version in the [Gradle Plugin Portal](https://plugins.gradle.org/plugin/co.elastic.otel.android.instrumentation.crash).
```
alexandra5000:
```suggestion
Find the latest crash reporting plugin version in the [Gradle Plugin Portal](https://plugins.gradle.org/plugin/co.elastic.otel.android.instrumentation.crash).
```

Numbered lists only make sense if there are at least 2 steps that are done in that order :) Otherwise there's no point

alexandra5000 (commented):
Nicely written, thanks @LikeTheSalad! I left some comments, feel free to ping me for approval when you apply them or wait for @vishaangelova, whichever you prefer 🙂

docs/reference/edot-android/crash-reporting.md
LikeTheSalad: I agree it's better to show users the min Stack version too; however, I'm concerned that by using `9.3.9`, it might cause users to think that this feature is not available in `8.x` versions. Is there a way to specify 2 min versions? (e.g. "8.19.20" and "9.3.9").

docs/reference/edot-android/configuration.md
LikeTheSalad: I think the solution for [this other thread](https://github.com/elastic/apm-agent-android/pull/877#discussion_r3776600789) might be applicable here too.

docs/reference/edot-android/crash-reporting.md
LikeTheSalad: If not, I think I'd like the other option you proposed of mentioning the content package link, as it contains all the min versions.

docs/reference/edot-android/crash-reporting.md
LikeTheSalad: Thanks! They do happen in order, so in this case it should stay numbered.

docs/reference/edot-android/crash-reporting.md
LikeTheSalad: It is, but they won't be able to install the version `1.0.0` of the content package if they don't have the right Kibana version, so I thought it might be redundant to mention the Kibana versions here as well. Though I think I'll add a link that shows the prerequisites of the content package, where it shows the supported Kibana versions, so that it's clearer for users and that way we avoid duplicating that info here. Wdyt?

jackshirazi (dismissed):

.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           api-level: ${{ matrix.api-level }}
           script: |
             make -C .github/scripts/integration-test run-tests
-            make -C .github/scripts/integration-test run-crash-test
```
mallendem, line 83: Are we ok with removing the `run-crash-test` target?

LikeTheSalad: We're technically not removing it but renaming it to `run-crash-deobfuscation-test` as we're expanding it to not only run crash validations but also crash deobfuscation ones. However, the name `run-crash-deobfuscation-test` is not clear enough, so I just renamed it to `run-crash-and-deobfuscation-test`.

docs/reference/edot-android/crash-reporting.md
alexandra5000: Well, if they won't be able to do it anyway, then it makes sense to skip it, but only if they users will get some sort of info why the install fails (if they attempt it). 

What I'd like to avoid is a situation when the installation fails or something else doesn't work because of the version, and there is no info in the docs or anywhere else explaining what happened.

docs/reference/edot-android/crash-reporting.md
alexandra5000: You know what, having slept on it I would also lean towards the link. We can remove the stack from tags then and just keep the link, especially since only specific versions are supported, not all of them since 8.19.20

docs/reference/edot-android/crash-reporting.md
LikeTheSalad: Got it. I added more details to this item to avoid confusion

docs/reference/edot-android/crash-reporting.md
LikeTheSalad: Got it, thanks! I've added the link with a description explaining to users that they can see Kibana-supported versions there. Cheers!

alexandra5000 (dismissed):

alexandra5000 (approved):

hegerchr (approved):
LGTM

mallendem (approved):

PR #909, opened 2026-09-14
Enhance the release process

LikeTheSalad:
## Summary

Previously, the release process required some manual steps that could be further automated, such as manually forming a set of release notes and passing them to the release preparation action, and also running the release workflow manually and ensuring that the correct release branch was selected. Now, a "prepare release" action dispatch opens a preparation pull request into a `releasing/x.y.z` branch, with an additional/optional path of using an AI skill to help with writing the release notes and also helping with other arguments needed to prepare the release. Once the generated PR gets merged, the actual release is executed automatically rather than having to manually continue with the process of finding the right next GH action and running it by choosing the right branch. The automation then opens a second pull request into `main` with the release changes and the next development version. Nothing is merged by automation.

What this helps with:

- The process makes the overall release less error-prone, as the version for releasing isn't chosen manually in any step, and it's also verified based on the latest released tag available to avoid accidental version skips.
- Buildkite builds exactly the selected commit and refuses anything else.
- A failed run is rerun in place, and a half-published release can continue where it left off.
- The outcome reaches Slack with links to the GitHub Release and the follow-up pull request, or to the failed run.
- Fewer moving parts to maintain: commented shell scripts replace the Kotlin renderer and the one-off scripts.
- An optional AI skill can help draft the initial release preparation PR. The skill won't do anything that cannot be done manually already; it just makes the process smoother.

## Changes

- Rewrite `prepare-release.yml` and its script: derive the version, render the release notes at the existing marker, regenerate the NOTICE files, push `releasing/x.y.z` and `prepare/x.y.z`, and open the preparation pull request.
- Add `publish-release.yml` with a guard and a finalize script: on merge into `releasing/x.y.z`, build and publish once in Buildkite from the merged commit, attest the artifacts, tag, create the GitHub Release, commit the next `-SNAPSHOT` version, and open the release pull request into `main`.
- Rewrite `.ci/release.sh` to build only the merged commit, verify credentials before any upload, publish per target, and check the registries only when an upload fails so a rerun finishes a partial release; add a root Gradle task listing the published Gradle plugins.
- Add the `draft-release-notes.yml` and `release-dry-run.yml` workflows, the notes draft, render, PR range, and version scripts, the `android-release-wizard` skill, dependency labels for Renovate and Dependabot, and the `1.10.0-SNAPSHOT` development version; rewrite `RELEASING.md`.
- Replace `release.yml`, `draft-changelog.yml`, the Buildkite dry run in `ci.yml`, and the previous release scripts.


.github/workflows/prepare-release.yml
```diff
@@ -45,75 +51,15 @@ jobs:
       - uses: actions/checkout@v7
         with:
           fetch-depth: 0
-          token: ${{ steps.get_token.outputs.token }}
+          token: ${{ steps.token.outputs.token }}
```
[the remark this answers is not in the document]
LikeTheSalad: Not needed: with no ref input, actions/checkout fetches the triggering github.sha into refs/remotes/origin/main and checks out from there, so the working tree is already pinned to the dispatched commit even if main advances while the run is queued. The script's GITHUB_SHA and the checkout agree.

.ci/release.sh
```diff
@@ -1,44 +1,149 @@
...
+  echo "--- Verify the publishing credentials"
+  # Stop before uploading anything if a credential for a selected target is
+  # missing, so a missing Portal secret cannot leave Maven Central published
+  # and the Portal not. The Vanniktech publish plugin reads the
+  # ORG_GRADLE_PROJECT_* variables as Gradle properties.
+  if [[ $target_specifier == all || $target_specifier == mavenCentral ]]; then
+    : "${ORG_GRADLE_PROJECT_mavenCentralUsername:?missing Maven Central credential}"
+    : "${ORG_GRADLE_PROJECT_mavenCentralPassword:?missing Maven Central credential}"
+    : "${ORG_GRADLE_PROJECT_signingKey:?missing Maven Central signing key}"
+    : "${ORG_GRADLE_PROJECT_signingPassword:?missing Maven Central signing password}"
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,138 @@
...
+concurrency:
+  group: publish-release-${{ github.event.pull_request.base.ref }}
+  cancel-in-progress: false
+
+jobs:
+  publish-release:
+    name: Publish release
+    if: >-
+      github.event.pull_request.merged == true
+      && startsWith(github.event.pull_request.base.ref, 'releasing/')
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.github/scripts/release/finalize-release.sh
```diff
@@ -0,0 +1,128 @@
...
+  echo "$base_ref has version $branch_version; expected $release_version or $next_development." >&2
+  exit 1
+fi
+
+# The PR into main carries the release changes and the version bump. An
+# operator merges it; the bot never does.
+main_pr=$(
+  gh pr list \
+    --repo "$repository" \
+    --state all \
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.github/workflows/prepare-release.yml
```diff
@@ -5,31 +5,37 @@ on:
...
+        options:
+          - minor
+          - major
 
 permissions:
   contents: read
 
 jobs:
-  prepare_release:
+  prepare-release:
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

skills/android-release-wizard/SKILL.md
```diff
@@ -0,0 +1,80 @@
...
+proposal shown in that summary. Treat a question, a comment, partial
+feedback, or silence as more feedback. When unsure whether the operator
+approved, ask.
+
+## Dispatch
+
+1. Show the exact command with the approved inputs substituted, and run it:
+
+   ```sh
+   gh workflow run prepare-release.yml --ref main -f release_notes='<json>' -f bump='<minor-or-major>'
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.github/scripts/release/render-release-notes.sh
```diff
@@ -0,0 +1,96 @@
...
+  and (.fixes | type == "array")
+  and (.uncategorized | type == "array")
+  and ([.dependencies[], .featuresEnhancements[], .fixes[], .uncategorized[]]
+    | all(
+        type == "object"
+        and (.message | type == "string" and length > 0 and (test("[\\r\\n]") | not))
+        and (
+          .prId == null
+          or (.prId | type == "number")
+          or (.prId | type == "string" and test("^[0-9]+$"))
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

skills/android-release-wizard/SKILL.md
```diff
@@ -0,0 +1,80 @@
...
+   ```
+
+2. Report the run URL:
+
+   ```sh
+   gh run list --workflow prepare-release.yml --limit 1 --json url --jq '.[0].url'
+   ```
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.ci/release.sh
```diff
@@ -1,44 +1,150 @@
...
+: "${release_sha:?release_sha is required}"
+: "${target_specifier:?target_specifier is required}"
+: "${dry_run:?dry_run is required}"
+release_version=${release_version:-$(sed -n 's/^version=//p' gradle.properties)}
+
+echo "--- Verify the checked-out commit"
+# The workflow asks Buildkite to build one specific commit: the merge commit
+# of the preparation PR. Refuse anything else, so a branch that moved after
+# the merge can never be released.
+if [[ $(git rev-parse HEAD) != "$release_sha" ]]; then
```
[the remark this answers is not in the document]
LikeTheSalad: Nothing in the current build reads the Git branch or HEAD state: build-tools has no release or versioning plugin, and the only HEAD is the literal string written into the POM's scm.tag. The removed checkout and its comment date from 2023 and a different toolchain. The release dry run runs this same Buildkite checkout on every push to main, so the first run after merge will confirm it. If Gradle fails on a detached HEAD there, I will attach a branch to the merged SHA and keep the assertion.

.github/scripts/release/publish-guard.sh
```diff
@@ -0,0 +1,68 @@
...
+if [[ $# -ne 3 ]]; then
+  echo "Usage: publish-guard.sh <base-ref> <head-ref> <merge-commit>" >&2
+  exit 2
+fi
+
+base_ref=$1
+head_ref=$2
+release_sha=$3
+
+git fetch --quiet --tags origin "$release_sha"
```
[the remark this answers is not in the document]
LikeTheSalad: Fetching by bare SHA works against GitHub; I ran the same fetch against this repository and actions/checkout relies on the same behavior. The job also checks out the base branch with full history, whose tip is this merge commit, so the object is already local. Leaving as is.

.github/scripts/release/render-release-notes.sh
```diff
@@ -0,0 +1,96 @@
...
+  and (.fixes | type == "array")
+  and (.uncategorized | type == "array")
+  and ([.dependencies[], .featuresEnhancements[], .fixes[], .uncategorized[]]
+    | all(
+        type == "object"
+        and (.message | type == "string" and length > 0 and (test("[\\r\\n]") | not))
+        and (
+          .prId == null
+          or (.prId | type == "number" and . == floor and . > 0)
+          or (.prId | type == "string" and test("^[1-9][0-9]*$"))
```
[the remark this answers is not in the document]
LikeTheSalad: jq's and short-circuits. Tested: '"901" | (type == "number" and . == floor and . > 0)' returns false without evaluating floor, and 'false and error("x")' returns false. The sample with string ids and a case with a numeric id both render correctly.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,140 @@
...
+      && startsWith(github.event.pull_request.head.ref, 'prepare/')
```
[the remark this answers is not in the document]
LikeTheSalad: A fork PR cannot reach publication: GitHub withholds secrets from runs triggered by fork pull requests, so the App token cannot be minted and the job fails before the guard. A same-repository branch pushed by a write-access user is within the trust boundary this flow accepts. Leaving as is.

.github/scripts/release/prepare-release.sh
```diff
@@ -0,0 +1,148 @@
...
+prepare_branch="prepare/$release_version"
+
+# Only one release can be in flight. A releasing branch exists from the
+# moment preparation pushes it until the release PR into main is merged and
+# the branch deleted.
+existing_release_branches=$(git ls-remote --heads origin 'refs/heads/releasing/*' | awk '{print $2}')
+if [[ -n $existing_release_branches ]]; then
+  echo "A release is already in flight. Finish it, or delete its branches if it was abandoned:" >&2
+  sed 's|refs/heads/|  |' <<<"$existing_release_branches" >&2
+  exit 1
```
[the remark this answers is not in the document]
LikeTheSalad: Valid but intentionally handled by a person: the window is the prepare push or PR creation failing right after the releasing branch was pushed. The script's message names the branches to delete, and the recovery is the same as for changing a preparation: delete both branches and dispatch again. I documented that case in RELEASING.md rather than adding resume logic to the script.

.ci/release.sh
```diff
@@ -1,44 +1,150 @@
...
+# an existing version. The script therefore publishes first and, when an
+# upload fails, checks the registry to tell "already published" from a real
+# failure. That way a rerun after a partial failure finishes the release
+# instead of failing on what already went through.
+
+set -euo pipefail
+
+: "${release_sha:?release_sha is required}"
+: "${target_specifier:?target_specifier is required}"
+: "${dry_run:?dry_run is required}"
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.github/scripts/release/finalize-release.sh
```diff
@@ -0,0 +1,133 @@
...
+# already did it.
+next_development=$("$script_dir/version.sh" next-development "$release_version")
+git switch -C "$base_ref" "origin/$base_ref"
+branch_version=$(sed -n 's/^version=//p' gradle.properties)
+if [[ $branch_version == "$release_version" ]]; then
+  sed -i.bak "s/^version=.*/version=$next_development/" gradle.properties
+  rm gradle.properties.bak
+  git add -- gradle.properties
+  git commit -m "Prepare for the next release"
+  git push origin "$base_ref"
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,140 @@
...
+              "contents": "write",
+              "pull_requests": "write"
+            }
+          repositories: >-
+            ["apm-agent-android"]
+
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.base.ref }}
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,140 @@
...
+
+      - name: Guard merged release
+        id: guard
+        env:
+          GH_TOKEN: ${{ steps.token.outputs.token }}
+        run: |
+          .github/scripts/release/publish-guard.sh \
+            "${{ github.event.pull_request.base.ref }}" \
+            "${{ github.event.pull_request.head.ref }}" \
+            "${{ github.event.pull_request.merge_commit_sha }}"
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,140 @@
...
+      - name: Finalize release
+        id: finalize
+        env:
+          GH_TOKEN: ${{ steps.token.outputs.token }}
+        run: |
+          .github/scripts/release/finalize-release.sh \
+            "${{ steps.guard.outputs.release_sha }}" \
+            "${{ steps.guard.outputs.release_version }}" \
+            "${{ github.event.pull_request.base.ref }}" \
+            "${{ steps.guard.outputs.publish_required }}"
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes.

.github/workflows/prepare-release.yml
```diff
@@ -5,31 +5,43 @@ on:
...
         required: true
         type: string
+      bump:
+        description: Release version bump.
+        required: true
+        default: minor
+        type: choice
+        options:
+          - minor
+          - major
```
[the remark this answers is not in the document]
LikeTheSalad: Patch releases are intentionally out of scope for this change and will be handled separately. The previous patch path was manual work on a release branch, not something this change removes from the automation.

.ci/release.sh
```diff
@@ -1,44 +1,167 @@
...
     ./gradlew assemble
+  else
+    echo "--- Release the binaries to Maven Central"
+    if ! ./gradlew publishAndReleaseElasticToMavenCentral "${common_gradle_deploy_params[@]}"; then
+      # All modules go to Maven Central in one deployment that is released or
+      # dropped as a whole, so one module's POM tells whether this version is
+      # already public.
+      central_pom="https://repo1.maven.org/maven2/co/elastic/otel/android/agent-sdk/$release_version/agent-sdk-$release_version.pom"
+      if is_public "$central_pom"; then
+        echo "Maven Central already has $release_version; nothing to publish."
```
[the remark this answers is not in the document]
LikeTheSalad: The per-subproject task registration does not split the upload. The publishing plugin's build service collects every publication of one Gradle invocation into a single Central Portal deployment, uploads it as one bundle at the end of the build, and releases or drops it as a whole, so all modules are public or none is. One module's POM therefore answers for the deployment. Leaving as is.

.github/scripts/release/finalize-release.sh
```diff
@@ -0,0 +1,138 @@
...
+if [[ $branch_tip == "$release_sha" ]]; then
+  git switch -C "$base_ref" "$release_sha"
+  sed -i.bak "s/^version=.*/version=$next_development/" gradle.properties
+  rm gradle.properties.bak
+  git add -- gradle.properties
+  git commit -m "Prepare for the next release"
+  git push origin "$base_ref"
+elif [[ $(git rev-parse "$branch_tip^") == "$release_sha" \
+  && $(git show "$branch_tip:gradle.properties" | sed -n 's/^version=//p') == "$next_development" ]]; then
+  echo "$base_ref already carries the next development version."
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes: the previous bump is accepted only if the diff from the merged commit touches gradle.properties alone.

.github/scripts/release/finalize-release.sh
```diff
@@ -0,0 +1,141 @@
...
+fi
+
+# Move the release branch to the next development version. The tag stays on
+# the merged commit; this commit comes right after it. The branch must still
+# point at the merged commit, or at the bump commit from a previous run of
+# this script: anything else means something was pushed after the merge and
+# would reach main without having been published, so stop.
+next_development=$("$script_dir/version.sh" next-development "$release_version")
+branch_tip=$(git rev-parse "origin/$base_ref")
+if [[ $branch_tip == "$release_sha" ]]; then
```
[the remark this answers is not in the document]
LikeTheSalad: The order is deliberate. By the time this script runs, the artifacts are already public on Maven Central and the Plugin Portal; that is the irreversible step. The tag and the GitHub Release describe those artifacts at the commit that produced them and are correct whatever the branch did afterwards. Creating them before the branch check means a rerun can skip republication through the tag lock. Stopping before them would leave published artifacts without a tag or Release, which is the worse state. Recovery for a moved branch is to remove the stray commit and rerun; the tag, Release, and publication are then skipped and only the bump and release PR proceed.

.ci/release.sh
```diff
@@ -1,44 +1,167 @@
...
+# upload fails, checks the registry to tell "already published" from a real
+# failure. That way a rerun after a partial failure finishes the release
+# instead of failing on what already went through.
+
+set -euo pipefail
+
+: "${release_sha:?release_sha is required}"
+: "${target_specifier:?target_specifier is required}"
+: "${dry_run:?dry_run is required}"
+release_version=${release_version:-$(sed -n 's/^version=//p' gradle.properties)}
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes: a real publish now requires release_version to be X.Y.Z and to equal the version in gradle.properties at the checked-out commit.

.github/scripts/release/prepare-release.sh
```diff
@@ -0,0 +1,148 @@
...
+fi
+
+# Everything checked out. From here on the script writes to GitHub.
+# The base branch is the dispatched commit, untouched; the release changes
+# go on the prepare branch so the PR diff shows exactly what the release
+# adds.
+git branch "$release_branch" "$release_ref"
+git push origin "refs/heads/$release_branch"
+git switch -c "$prepare_branch"
+git add -u
```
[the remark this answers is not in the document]
LikeTheSalad: Valid. I applied the changes: the preparation commit now stages additions for the root NOTICE and every META-INF/NOTICE, together with gradle.properties and docs, so a module that gains its first NOTICE ships it.

LikeTheSalad:
Quoted from the last [Copilot message](https://github.com/elastic/apm-agent-android/pull/909#pullrequestreview-5202508044)

> This only filters the new releasing/* and prepare/* bookkeeping branches. The previous workflow in this PR created post-release/<version> (and release/*/pre-release/*) PRs, and the post-release version-bump PR is after the previous release tag, so the first draft will include that bookkeeping as a release change.

This is as intended. It's basically saying that, since the new process uses a new set of branch name patterns different from the old one, that means that it will take the last release PRs as release notes candidates (since the filter that excludes PRs for release notes only takes into account the new release branch names). However, this will only happen once (in the next release), and I can manually exclude those by then, so I don't like the idea of adding a validation for a one-off use case.

.ci/release.sh
```diff
@@ -1,44 +1,194 @@
...
+# an existing version. The script therefore publishes first and, when an
+# upload fails, checks the registry to tell "already published" from a real
+# failure. That way a rerun after a partial failure finishes the release
+# instead of failing on what already went through.
+
+set -euo pipefail
+
+: "${release_sha:?release_sha is required}"
+: "${target_specifier:?target_specifier is required}"
+: "${dry_run:?dry_run is required}"
```
v1v: wonder if this should be defaulted to 'true', rather than enforcing it?

.github/scripts/release/draft-release-notes.sh
```diff
@@ -0,0 +1,66 @@
...
+# lands in `uncategorized` for the operator to place or delete. Repeated
+# updates of the same dependency collapse into the last one merged. The output
+# is a starting point; the operator edits it before dispatching.
+
+set -euo pipefail
+
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+range=$("$script_dir/pr-range.sh" "${1:-HEAD}" "${2:-}")
+
+jq 'def has_label($name):
```
v1v, line 20: Although I'm not against, this is a bit too much, wonder what's the consequence if this script does not work as expected? If it's important, what if moving this logic to a gradle function or some other technology that will provide some test coverage?

.github/scripts/release/finalize-release.sh
```diff
@@ -0,0 +1,141 @@
...
+  echo "Usage: finalize-release.sh <release-sha> <release-version> <base-ref> <create-tag:true|false>" >&2
+  exit 2
+fi
+
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+release_sha=$1
+release_version=$2
+base_ref=$3
+create_tag=$4
+repository=${GITHUB_REPOSITORY:?GITHUB_REPOSITORY is required}
```
v1v, line 43: IIUC, this is a requirements, wonder if we shall update the comments with the list of required env variables and what's the format of their value?

.github/scripts/release/finalize-release.sh
```diff
@@ -0,0 +1,141 @@
...
+if [[ $# -ne 4 ]]; then
+  echo "Usage: finalize-release.sh <release-sha> <release-version> <base-ref> <create-tag:true|false>" >&2
+  exit 2
+fi
+
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+release_sha=$1
+release_version=$2
+base_ref=$3
+create_tag=$4
```
v1v: fail if those values are empty?

v1v (commented):
I left a few comments, but didn't finish the whole review, but wanna provide a quick feedback here.

.github/scripts/release/draft-release-notes.sh
LikeTheSalad: > what's the consequence if this script does not work as expected?

It's used to provide a draft only, which someone (or an AI, or both) later has to double-check to make sure that the items collected should actually make it to the release notes page. If it were producing final release notes content, then I agree we should be more careful about its output, but it's more of a help to avoid manually finding and categorizing all the PRs that got merged since the last release.

Now, this was written by an AI, and I agree that putting all this script inline here seems like a lot. I probably wouldn't have added a jq script and instead would rely on some shell functions, but that would probably be a larger script in the end. I still can try and ask the agent to do so, at least to have a more readable version of it; wdyt?

.github/scripts/release/finalize-release.sh
LikeTheSalad: Thanks! I added the changes.

.github/scripts/release/finalize-release.sh
LikeTheSalad: Good call. I added the validations.

.ci/release.sh
LikeTheSalad: Sounds way safer, thank you. I added the changes.

skills/android-release-wizard/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+
+- `release_notes`: JSON with `dependencies`, `featuresEnhancements`, `fixes`,
+  and `uncategorized` arrays of `{message, prId}` items. Prepare release
+  rejects leftover `uncategorized` items and empty notes.
+- `bump`: `minor`, or `major` when the release contains a breaking change.
+  A breaking item's message starts with `[Breaking]`.
+
+## Propose
+
+1. Run from the repository root:
```
vishaangelova: Shouldn't we make sure it does a `git fetch` first?

RELEASING.md
```diff
@@ -1,69 +1,124 @@
...
-* A compilation issue. There are several reasons and compilation checks that can make a build fail. Refer to the [build-tools's README file](build-tools/README.md) to get a comprehensive guide on the compilation details of this project.
+A release takes three steps: dispatch the preparation, merge the preparation
+PR, merge the release PR. The version is derived from the highest release
+tag. Merging the preparation PR is what publishes.
+
+### 1. Prepare the release
+
+Use either path:
+
+- Ask an agent to use the `android-release-wizard` skill.
```
vishaangelova: Maybe we could say upfront what happens when the agent uses the skill?

```suggestion
- Ask an agent to prepare the release with the
  [`android-release-wizard`](skills/android-release-wizard/SKILL.md)
  skill. It proposes the notes and bump, then dispatches after you
  approve.
```

vishaangelova (dismissed):
LGTM for the docs

skills/android-release-wizard/SKILL.md
LikeTheSalad: I think this makes a lot of sense to ensure the skill is up-to-date, that's a great point, I'll add the changes.

RELEASING.md
LikeTheSalad: Sounds good! I'll add the changes.

skills/android-release-wizard/SKILL.md
LikeTheSalad: I added some instructions to ensure that the local repo is updated before starting, but I prefer not to add an explicit instruction on how to update it so that the AI agent chooses what's best in that case. For example, if I added "run 'git fetch origin'", that might not work in cases where the git remote isn't called "origin". I also noticed some similar issues in the skill and addressed them as well. Thanks!

RELEASING.md
LikeTheSalad: It's added now, thanks!

vishaangelova (approved):
LGTM! 🚀

.github/scripts/release/draft-release-notes.sh
v1v: Thanks for the explanation here, then I think we don't need to refactor this

.ci/release.sh
```diff
@@ -1,44 +1,196 @@
...
+
+# True when the URL answers HTTP 200 to a HEAD request. Only 200 counts: the
+# Gradle Plugin Portal answers a missing version with a 303 redirect, which
+# `curl -f` would accept as success.
+is_public() {
+  [[ $(curl -s -o /dev/null -w '%{http_code}' -I "$1") == 200 ]]
+}
+
+if [[ $target_specifier == all || $target_specifier == mavenCentral ]]; then
+  if [[ $dry_run == true ]]; then
```
v1v: `nit`:

I'd prefer if we use `if [[ $dry_run == false ]]; then` instead.

**why?** because I think `dry_run` can be anything, and mixing the conditional with `true` validations or else, or false validation and else in the same script can be misleading and allow some misbehaviours if by accident we run the script with the wrong value format.

.ci/release.sh
```diff
@@ -1,44 +1,196 @@
...
-    echo "--- Release the binaries to the Gradle Plugin portal"
-    ./gradlew publishPlugins -Pgradle.publish.key=$PLUGIN_PORTAL_KEY -Pgradle.publish.secret=$PLUGIN_PORTAL_SECRET $COMMON_GRADLE_DEPLOY_PARAMS
-  else
-    echo "--- Release the binaries to Gradle Plugin portal :package: (dry-run)"
+if [[ $target_specifier == all || $target_specifier == pluginPortal ]]; then
+  if [[ $dry_run == true ]]; then
+    echo "--- Release the binaries to the Gradle Plugin Portal :package: (dry-run)"
     ./gradlew assemble
+  else
+    echo "--- Release the binaries to the Gradle Plugin Portal"
```
v1v: `nit`:

wonder if we can create reusable shell functions and simplify the main function to call them, for instance:


```shell

# run gradle assemble without running any extra step
# help function when running with dry_run=true
function release_dry_run () {
  echo "--- Release the binaries to the Gradle Plugin Portal :package: (dry-run)"
  ./gradlew assemble
}

# run the release for the gradle portal
# ...
function release_gradle_portal () {
  echo "--- Release the binaries to the Gradle Plugin Portal"
  ...
}

if [[ $target_specifier == all || $target_specifier == pluginPortal ]]; then
  if [[ $dry_run == true ]]; then
    release_dry_run
  else
    release_gradle_portal
  fi
fi
```

.github/scripts/release/pr-range.sh
```diff
@@ -0,0 +1,93 @@
...
+#
+# Usage: pr-range.sh [ref] [previous-tag]
+#
+# Walks the first-parent commits between the previous release tag and `ref`
+# (default HEAD) and resolves each one to its merged pull request through the
+# GitHub API. Because `main` is squash-merged, each commit is one PR. A commit
+# with no PR stops the script: the release notes would otherwise miss a
+# change. PRs from `releasing/*` and `prepare/*` branches are release
+# bookkeeping and are left out. Prints JSON with the tag, the range, and one
+# entry per PR (number, title, URL, labels).
```
v1v, line 23: missing the same header comment with the arguments and environments, something like:

but I didn't fill the description for the arguments

```suggestion
# entry per PR (number, title, URL, labels).
#
# Arguments:
#   ref                   ...
#   previous_tag  ...
#
# Environment:
#   GITHUB_REPOSITORY  owner/repo of the release repository (required)
#   GH_TOKEN           GitHub App token with contents
#                      read access, used by every gh call (required)
```

.github/scripts/release/prepare-release.sh
```diff
@@ -0,0 +1,150 @@
...
+# after checking that gradle.properties holds the expected `-SNAPSHOT`
+# version. Then pushes `releasing/X.Y.Z`, an unchanged copy of the
+# dispatched commit that serves as the PR base, and commits once on
+# `prepare/X.Y.Z`: the version, the rendered release-notes section, the
+# documentation `applies_to` versions on a major bump, and the regenerated
+# NOTICE files. Opens the PR from `prepare/X.Y.Z` into `releasing/X.Y.Z`.
+# Merging that PR publishes the release.
+#
+# All checks run before anything is pushed. If no PR was merged since the
+# last release, the script exits successfully without creating anything.
```
v1v, line 27: missing the same header comment with the arguments and environments, something like:

.github/scripts/release/publish-guard.sh
```diff
@@ -0,0 +1,68 @@
...
+# irreversible happens:
+#   - the merged commit is a prepared release: gradle.properties at the merge
+#     commit holds a release version X.Y.Z, the base branch is
+#     `releasing/X.Y.Z`, and the merged head is `prepare/X.Y.Z`, so no other
+#     PR into the releasing branch can publish;
+#   - whether it was already published: if tag vX.Y.Z exists it must point at
+#     the merge commit, in which case a previous run published and failed
+#     later, publication is skipped, and only the finalization steps run. A
+#     tag at another commit stops the run.
+# Outputs the release facts for the following workflow steps.
```
v1v, line 28: missing the same header comment with the arguments and environments, something like:

.github/scripts/release/render-release-notes.sh
```diff
@@ -0,0 +1,99 @@
...
+# docs/release-notes/index.md.
+#
+# Usage: render-release-notes.sh <source-json> <version>
+#
+# Validates the JSON shape, refuses input that still has `uncategorized`
+# items or has no items, and prints a `## X.Y.Z` section with the anchors the
+# docs site expects: an untitled list for dependencies, then "Features and
+# enhancements" and "Fixes" subsections. Empty groups are omitted and
+# `[Breaking]` items sort first. The same section becomes the GitHub Release
+# body.
```
v1v, line 21: missing the same header comment with the arguments and environments, something like:

.github/scripts/release/render-release-notes.sh
```diff
@@ -0,0 +1,99 @@
...
+if [[ $# -ne 2 ]]; then
+  echo "Usage: render-release-notes.sh <source-json> <version>" >&2
+  exit 2
+fi
+
+source_file=$1
+version=$2
+release_date=$(date +'%B %d, %Y' | sed 's/ 0/ /')
+repository=${GITHUB_REPOSITORY:-elastic/apm-agent-android}
+server_url=${GITHUB_SERVER_URL:-https://github.com}
```
v1v: what's the reason for allowing configuring a different `GITHUB_SERVER_URL`?

.github/workflows/draft-release-notes.yml
```diff
@@ -0,0 +1,31 @@
...
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  draft:
+    runs-on: ubuntu-latest
```
v1v:
```suggestion
    runs-on: ubuntu-slim
```

.github/workflows/prepare-release.yml
```diff
@@ -5,31 +5,43 @@ on:
...
-            exit 1
-          fi
+          echo "::error::Prepare release must be dispatched from main."
+          exit 1
 
-      - name: Get token
-        id: get_token
+      - name: Get automation token
+        id: token
         uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
```
v1v: let's use https://github.com/actions/create-github-app-token, instead `tibdex/github-app-token`

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,142 @@
...
+      attestations: write
+      contents: write
+      id-token: write
+      pull-requests: write
+    env:
+      TARBALL_FILE: dist.tar
+    steps:
+      - name: Get automation token
+        id: token
+        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
```
v1v: let's use https://github.com/actions/create-github-app-token, instead `tibdex/github-app-token`

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,142 @@
+---
+name: Publish release
+
+# Runs when a preparation PR is merged into its releasing/x.y.z branch.
+on:
+  pull_request:
+    types: [closed]
```
v1v, line 7: wonder if you can add the filter `branches` here:
- https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-the-head-or-base-branch-of-a-pull-request


```suggestion
    types: [closed]
    branches:
      - 'releasing/**'
```

.github/workflows/release-dry-run.yml
```diff
@@ -0,0 +1,68 @@
+---
+name: Release dry run
```
v1v, line 2: What's the reason for moving this step from the `ci.yml`?

v1v (commented):
Just left a few minor comments

.github/workflows/release-dry-run.yml
LikeTheSalad: It’s mostly to keep the release workflow simpler and friendlier, since someone unfamiliar with the dry-run test might be confused by that option. But we can move it back into CI if that’s better for you.

.ci/release.sh
LikeTheSalad: Thanks! I'll make the changes

.ci/release.sh
LikeTheSalad: Thanks! I’ll split the dry-run and target-specific publication paths into functions while preserving the existing retry and error handling.

.github/scripts/release/pr-range.sh
LikeTheSalad: Thanks! I’ll add explicit documentation for the arguments and environment.

.github/scripts/release/prepare-release.sh
LikeTheSalad: Thanks! I’ll document the remaining environment inputs.

.github/scripts/release/publish-guard.sh
LikeTheSalad: Thanks! I’ll document the arguments and output environment.

.github/scripts/release/render-release-notes.sh
LikeTheSalad: Thanks! I’ll document the arguments and environment.

.github/scripts/release/render-release-notes.sh
LikeTheSalad: Thanks! There’s no expected use for a different GitHub host here, so I’ll simplify this and use `https://github.com` directly.

.github/workflows/prepare-release.yml
LikeTheSalad: Thanks! I’ll switch this to `actions/create-github-app-token`.

.github/workflows/publish-release.yml
LikeTheSalad: Thanks! I’ll switch this to `actions/create-github-app-token`.

.github/workflows/publish-release.yml
LikeTheSalad: Thanks! I'll add the branch filter

.github/workflows/release-dry-run.yml
v1v: Probably I'm biased here and it's just cosmetic.

Thanks for the explanation, no need to revert what you did

vishaangelova (approved):

v1v (approved):

